PR #13, opened 2026-07-12
Wan logging controls

lucas-albers-lz4:


lucas-albers-lz4:
Passes manual test on 24.10

PR #27, opened 2026-07-23
refactor: Phase 3 extract fwlive table (#23)

lucas-albers-lz4:
## Summary
- Add `fwlive/table.js` with locked APIs `renderThead` / `renderRows` (internals: column cells, flow).
- View keeps orchestration, scrollTop save/restore, empty/chips/status/tint probe.
- Entry `wc -l`: 1432 (guideline ~600–800 orchestration — soft; remaining LOC is poll/prefs/logging glue).
- `table.js`: 228 LOC. Part of #23.

## Test plan
- [x] `./scripts/fwlive-test.sh` + `./scripts/validate-baseline.sh`
- [ ] QEMU install + smoke
- [ ] scrollTop save/restore when paused / scrolled up
- [ ] Detail toggle thead/columns; row expand in simple mode
- [ ] Filter links in cells; hostnames; row tint
- [ ] 10× re-render sanity (DevTools: toggle detail / filters)

Made with [Cursor](https://cursor.com)

Issue #23, opened 2026-07-23
Design proposal: split fwlive.js modules (#17 follow-up)


lucas-albers-lz4:
## Context

Follow-up to [#17](https://github.com/lucas-albers-lz4/fwlive/issues/17) and its multi-model design review (REQUEST_CHANGES). That review correctly blocked implementation: the original #17 module table was underspecified on CSS loading and on how extracted modules access view state/callbacks.

This issue is the **design lock** before code moves. It is meant to be reviewable without reading the prior thread.

**Parent:** #17  
**Code under discussion:** [`view/status/fwlive.js`](openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js) (~2278 LOC today)  
**Existing helper:** [`fwlive/log.js`](openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/fwlive/log.js) (`baseclass.extend`, already `require`d)

---

## Problem we are solving

`fwlive.js` is a single LuCI `view.extend` file that mixes:

- Inline CSS (`E('style', {}, \`...\`)` inside `render()`, ~lines 1753–~2140)
- Module-scope constants and pure tint helpers (~57–124)
- Filter chip UI, WAN logging toolbar / empty state, table/row builders
- Poll/render loop, hash filters, localStorage prefs, hostname resolve, stream controls

That makes review, upstream iteration, and targeted tests harder than peer LuCI apps that keep helpers under `resources/`.

**Non-goals (unchanged from #17):**

- No user-visible behavior change (filters, hash, localStorage, poll caps, tint probe, logging enable/disable)
- No rpcd / ACL / shell changes
- No UX redesign (e.g. chip colors — separate issue)
- Not a merge blocker for upstream; maintainability debt only

---

### A. CSS extraction mechanism

| Option | Mechanism | Pros | Cons |
|--------|-----------|------|------|
| **A1. `fwlive/css.js`** | Export `{ styleText }`; view still emits `E('style', {}, css.styleText)` | Matches existing `require fwlive.*` pattern; same FOUC/CSP behavior as today; no Makefile/link inventiveness | Still an inline `<style>` node (not a cacheable `.css` URL) |
| **A2. `fwlive/style.css`** | Ship static CSS + inject `<link>` | “Cleaner” static asset in theory | LuCI view packaging/load path for app-local CSS is non-obvious here; FOUC/load-order risk; more packaging surface |

[...]

PR #28, opened 2026-07-24
fix: fwlive modules must use baseclass.extend (#23)

lucas-albers-lz4:
## Summary
- LuCI `require()` only accepts `baseclass.extend(...)` factories; plain `return {…}` raises `factory yields invalid constructor`.
- Wrap `constants`, `css`, `tint`, `links`, `chips`, `logging`, and `table` the same way as `log.js`.
- Update host CSS/tint unit tests to stub `baseclass.extend`.

Fixes the live LuCI page error after #24–#27.

## Test plan
- [x] `./scripts/fwlive-test.sh`
- [x] Synced to QEMU 24.10.5; hard-refresh loads `#fwlive-table` with no invalid-constructor errors

Made with [Cursor](https://cursor.com)


lucas-albers-lz4:
Also fixes intermittent double-click on Enable/Disable logging: poll was wiping the button between mousedown and click. Synced to QEMU; 12/12 slow human-like clicks flipped mode on first press.

lucas-albers-lz4:
# Multi-Model Review — lucas-albers-lz4/fwlive PR #28

> **Head SHA:** `116860082cb37ba797e8f32dcb199ee8868bfb94`  •  **Reviewers:** opencode-go/qwen3.7-max, opencode-go/minimax-m3, opencode-go/kimi-k2.7-code  •  **Synthesized by:** opencode-go/deepseek-v4-pro

## TL;DR
This PR wraps seven fwlive LuCI modules in `baseclass.extend(...)` to fix the `factory yields invalid constructor` runtime error, and adds signature-based caching with `mousedown` handlers to fix an intermittent double-click bug on logging toggle buttons. The module-wrapping is clean and consistent. However, all three reviewers independently flag the `mousedown`-for-`click` swap as a keyboard accessibility regression — `<button>` elements listened on `mousedown` are unreachable via Enter/Space. The signature caching in `fwlive.js` likely solves the DOM-churn root cause on its own, making the `mousedown` workaround unnecessary (or at least requiring a companion `keydown` fallback). Additionally, `_loggingToolbarSig` / `_loggingEmptySig` are never declared in the data block, and several test-quality nits accumulate.

### 🔴 Critical
None.

### 🟡 Major

- **Finding:** `mousedown` replaces `click` on `<button>` elements, breaking keyboard + assistive-tech activation
- **File:** `openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/fwlive/logging.js:94-100`, `111-117`, `163-169`
- **Reviewers who flagged:** A, B, C (unanimous)
- **Problem:** Browsers dispatch `click` — not `mousedown` — when a focused button is activated via Enter or Space. Replacing `click` with `mousedown` makes the Enable/Disable logging buttons unreachable for keyboard-only users and screen-reader users, violating WCAG 2.1.1 (Keyboard). B further argues the `loggingUiSignature()` guard in `fwlive.js:549-561` already prevents the per-second DOM rebuild that caused the original double-click race, so `mousedown` is unnecessary in the steady state. A and C agree that even if `mousedown` is kept as a defense-in-depth measure, a `keydown` fallback on Enter/Space must be added.
- **Suggested fix:** (a) Restore `click` handlers since the signature cache already prevents per-second rebuilds; or (b) keep `mousedown` and add a `keydown` handler that fires on `Enter` / `Space` key codes for each button. A and C propose option (b); B prefers option (a).
- **Confidence:** high

### 🟢 Minor

- **Finding:** `_loggingToolbarSig` and `_loggingEmptySig` are written but never declared in the view’s data block
- **File:** `openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js:569, 571, 585, 587`
- **Reviewers who flagged:** B
- **Problem:** The two signature-cache keys are set on `this` but not initialized in the `data` section (contrast `loggingBusy: false`, `loggingNotice: ''` at `fwlive.js:94-95`). Functional since `undefined !== sig` holds, but inconsistent with the rest of the file and easy to miss during refactors.
- **Suggested fix:** Add `_loggingToolbarSig: ''` and `_loggingEmptySig: ''` to the return block’s data object.
- **Confidence:** high

- **Finding:** `blockers.join(',')` in `loggingUiSignature()` is order-sensitive — unstable backend order causes unnecessary rebuilds
- **File:** `openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js:551`
- **Reviewers who flagged:** C
- **Problem:** If the backend returns `blockers` in a different order across polls (e.g., `["a","b"]` vs `["b","a"]`), the signature changes and the toolbar/empty-state DOM is rebuilt even though nothing meaningful changed. This undermines the optimization the signature is meant to provide.
- **Suggested fix:** Sort before joining: `st.blockers.slice().sort().join(',')`.
- **Confidence:** high

- **Finding:** `entries.length` in `loggingUiSignature()` couples toolbar render to table row count
- **File:** `openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js:557`
- **Reviewers who flagged:** B (A mentions it as a nit, notes it’s already well-documented)
- **Problem:** The toolbar/empty-state DOM is rebuilt on the 0→1 and 1→0 entry-count transitions — which is the one transition where the original double-click bug can still occur (nodes are replaced while the user may be interacting). B argues the toolbar visuals don’t depend on entry count, so including it couples render cadence to table state unnecessarily.
- **Suggested fix:** Drop `entries.length` from the signature, or keep the bit but stop rebuilding the button subtree on that signal.
- **Confidence:** medium

- **Finding:** `preventDefault()` on `mousedown` suppresses button focus behavior
- **File:** `openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/fwlive/logging.js:98, 115, 167`
- **Reviewers who flagged:** A
- **Problem:** Calling `ev.preventDefault()` on `mousedown` prevents the browser’s default focus-on-click behavior — the button won’t receive focus after interaction, leading to missing focus rings and unexpected tab-order jumps.
- **Suggested fix:** Remove `preventDefault()` or explicitly call `ev.target.focus()` afterward.
- **Confidence:** medium

- **Finding:** Test baseclass stub returns raw descriptor object, not a constructor
- **File:** `tests/fwlive-theme-tint.test.js:30`, `tests/fwlive-theme-css.test.js:29`
- **Reviewers who flagged:** A
- **Problem:** The stub `{ extend: (desc) => desc }` returns the raw descriptor, while real LuCI `baseclass.extend()` returns a constructor whose instances carry those properties. Any future test that does `new Module()` or `instanceof` checks will silently diverge from production behavior.
- **Suggested fix:** Use a stub like `{ extend: (desc) => function() { Object.assign(this, desc); } }` and instantiate in tests.
- **Confidence:** low

### 💡 Nits

- **Finding:** Duplicated baseclass-stub + body-strip loader in two test files
- **File:** `tests/fwlive-theme-css.test.js:26-32`, `tests/fwlive-theme-tint.test.js:25-32`
- **Reviewers who flagged:** B
- **Suggested fix:** Extract a `loadFwliveModule(path)` helper into a shared test utility.

- **Finding:** New `mousedown` / button interaction path is not exercised by any host test
- **File:** `openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/fwlive/logging.js:88-170`
- **Reviewers who flagged:** B
- **Suggested fix:** Add a node-side test that loads `logging.js` with a stub `E`/`_` and asserts the button attributes.

- **Finding:** `'require baseclass';` is added to stateless helper modules with no class behavior
- **File:** `openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/fwlive/chips.js:2`, `links.js:2`
- **Reviewers who flagged:** B
- **Suggested fix:** Add a one-line comment explaining why a stateless function-module needs `baseclass.extend`.

- **Finding:** `ev.button` guard is dead code for `type='button'` elements
- **File:** `openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/fwlive/logging.js:95-96`
- **Reviewers who flagged:** B
- **Suggested fix:** Remove the `ev.button != null && ev.button !== 0` guard — browsers already ignore non-primary clicks on `<button type='button'>`.

- **Finding:** Redundant `ev && ev.preventDefault` truthy check
- **File:** `openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/fwlive/logging.js:98, 115, 167`
- **Reviewers who flagged:** B

- **Finding:** Docstring says `onEnable() → Promise` / `onDisable() → Promise` but callbacks are invoked synchronously
- **File:** `openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/fwlive/logging.js:12`
- **Reviewers who flagged:** C

- **Finding:** Unnecessary `m` flag in `replace(/^'require baseclass';\s*/m, '')`
- **File:** `tests/fwlive-theme-css.test.js:29`
- **Reviewers who flagged:** C

- **Finding:** Repeated explanatory comment `LuCI modules must return baseclass.extend(...)` in 3 of 7 modules
- **File:** `constants.js:6`, `css.js:6`, `tint.js:6`
- **Reviewers who flagged:** B

## Reviewer agreement matrix
|---|---|---|---|---|
| `mousedown` breaks keyboard accessibility | 🟡 | 🟡 | 🟡 | All three agree; A/C suggest adding keydown, B prefers restoring click |
| `_loggingToolbarSig`/`_loggingEmptySig` not declared | — | 🟢 | — | B only |
| `blockers.join(',')` order-sensitive | — | — | 🟢 | C only |
| `entries.length` in signature couples toolbar to table | 💡 | 🟢 | — | B flags as minor, A covers as nit |
| `preventDefault()` suppresses focus | 🟢 | — | — | A only |
| Test baseclass stub returns descriptor not constructor | 🟢 | — | — | A only |
| Duplicated test loader | — | 🟢 | — | B only |
| `mousedown` not tested | — | 🟢 | — | B only |
| Dead `ev.button` guard | — | 💡 | — | B only |
| Docstring `Promise` return type | — | — | 💡 | C only |

No substantive disagreements. All three agree the `mousedown` change is a keyboard accessibility regression and needs to be addressed. They differ only on the preferred fix: A and C suggest adding a `keydown` companion handler, while B argues the signature cache already solves the root cause and recommends reverting to `click` entirely. B also notes that even if `mousedown` is kept as defense-in-depth, it needs a `keydown` fallback — so the practical difference is minor.

None beyond the above. The module-wrapping change (`baseclass.extend` on all seven modules) is mechanically correct and consistent with the existing `log.js` pattern. The signature caching in `fwlive.js:549-561` is a well-structured solution covering all relevant state fields. The test updates correctly strip the new `'require baseclass';` line and inject a stub, and the `const vm = require('vm')` removal in the tint test is appropriate since the old sandbox code was already unused.

- [ ] Address the keyboard accessibility regression: either revert `mousedown` to `click` (if signature caching alone prevents the double-click race), or add `keydown` handlers on Enter/Space alongside `mousedown`
- [ ] Declare `_loggingToolbarSig: ''` and `_loggingEmptySig: ''` in the view’s data block for consistency
- [ ] Sort `blockers` before `join(',')` in `loggingUiSignature()` to avoid false signature changes from unstable backend ordering
- [ ] Consider dropping `entries.length` from the signature (toolbar visuals don’t depend on it) or document the intent
- [ ] Remove `preventDefault()` in the button handlers (or explicitly re-focus the target) if `mousedown` is retained
- [ ] Optionally: extract the baseclass-stub loader into a shared test helper, add a logging-button event test, clean up the dead `ev.button` guard

- Reviewer A (Qwen 3.7 Max): Thorough — covers module wrapping, UI interaction, testing, and WCAG implications. Identified the focus-suppression side-effect that B and C missed.
- Reviewer B (MiniMax M3): Most thorough — caught the missing data-block declarations, the duplicated test loader, the entries.length coupling, and multiple nits. Most critical of the `mousedown` approach, arguing it’s unnecessary given the signature cache.
- Reviewer C (Kimi K2.7-code): Adequate — correctly identifies the keyboard issue and the blockers ordering problem. Shorter report with fewer minor/nit findings than A or B.

## Process notes
- Wall time: 3 parallel reviews, ~18s each
- Reviewer failures: none
- Diff was truncated: no (27 KB, full 314-line patch)

lucas-albers-lz4:
Addressed multi-model REQUEST_CHANGES:

- **Major:** Reverted `mousedown` → `click` (keyboard Enter/Space works again). Signature cache alone prevents the per-second rebuild race (QEMU: 8/8 slow mouse + 4/4 Enter activations).
- **Minor:** Declared `_loggingToolbarSig` / `_loggingEmptySig` in the view data block.
- **Minor:** Sort `blockers` before `join`.
- **Minor:** Kept `entries.length` empty-bit in the signature — toolbar visibility depends on it (`!wan_log && no rows`); documented in-code.
- **Nits:** baseclass comment on chips/links; docstring no longer claims Promise return from sync invoke.

Skipped low-value test-stub constructor rewrite / shared loader extract for this PR.

lucas-albers-lz4:
Follow-up for the cumulative #23 review: dead wrapper cleanup, module smoke tests, signature-after-render, README/JSDoc. Finding #1 (mousedown a11y) was already fixed in the prior commit.

Issue #43, opened 2026-07-28
Multi-Model Codebase Review — design, security, and upstreaming readiness


lucas-albers-lz4:
# Multi-Model Codebase Review — lucas-albers-lz4/fwlive

> **Head SHA:** 6cf979844f54707162050af018cd3d674b193d93  •  **Reviewers:** Qwen 3.7 Max (A), MiniMax M3 (B), Kimi K2.7-code (C)  •  **Synthesized by:** DeepSeek V4 Pro
> **Overall assessment:** CONCERNING

## TL;DR

fwlive is a LuCI firewall Live View for OpenWrt — a live, filterable table of firewall LOG events. The codebase has strong security fundamentals (defense-in-depth filtering, tight ACL, validated DNS lookups, RFC 8259 JSON escaping) and the recent modularization PRs (#25–#28) were clean. However, a **data-loss bug on pause→resume** (found by B), a **likely poll-callback memory leak** (found by C), a **manually-mirrored parser that has already drifted** (found by A, B, C), and **browser-compat regressions on the 21.02 target** (found by A, C) collectively push the codebase into concerning territory. The root-running shell scripts also have multiple maintainability and latent correctness issues that warrant attention before upstreaming.

---

### 🔴 Critical

- **Finding:** Data loss on pause→resume — accumulated buffer silently replaced by new poll
- **File:** `htdocs/.../view/status/fwlive.js:889-895` + `fetchEntries` at `:626-630`
- **Category:** bug
- **Reviewers who flagged:** B
- **Problem:** When the user un-pauses, `trimEntriesToLiveCap` trims `this.entries` to `rowLimit`, then `fetchEntries()` runs and (because `!paused`) overwrites the trimmed buffer with `this.entries = normalized.slice(-ingestCap())`. The user loses up to 1900 events they accumulated while paused.
- **Suggested fix:** On the un-pause branch, route through `mergeEntries` (e.g., set a `resuming` flag) instead of the replace path in `fetchEntries`.
- **Confidence:** high

---

### 🟡 Major

- **Finding:** Parser mirror has already drifted — `actionRowClass` returns divergent class names
- **Files:** `core/fwlive-log.js:323-330` vs `htdocs/.../fwlive/log.js:397-404`
- **Category:** duplicate-code
- **Reviewers who flagged:** A, C (B flagged the duplicate-parser architecture broadly)
[...]

PR #63, opened 2026-07-29
fix(rpcd): nft timeout, rulesmap path lock, ubus hardening

### LINKED ISSUE #54: security: __rulesmap_iptables debug endpoint accepts arbitrary file paths
## Bug Description

In `root/usr/libexec/rpcd/fwlive:397-399`, the `__rulesmap_iptables` debug entry point accepts an arbitrary file path from its second argument with only `[ -f "$file" ]` validation:

```sh
__rulesmap_iptables)
    file="${2:-/tmp/rulesmap}"
    [ -f "$file" ] || return 1
```

Any user who can invoke `ubus call fwlive rules` (i.e., any LuCI-authenticated user with the `luci-app-fwlive` ACL, which includes all users on the status page) can pass an arbitrary path and have its contents parsed and partially returned through the iptables rules parser.

## Impact

While output is filtered through the iptables rule parser (not a raw file read), an attacker can probe for the existence and structure of arbitrary files (e.g., `/etc/shadow`, `/etc/config/*`).

## Suggested Fix

- Remove this debug entry point before wider distribution, or
- Whitelist allowed paths (e.g., only `/tmp/firewall*`)

**Found by:** MiMo V2.5 Free in Zen MCR. Confidence: high.

### LINKED ISSUE #61: bug: nft list ruleset called with no timeout — hangs indefinitely on corrupted ruleset
## Bug Description

In `rpcd/fwlive:159-169`, `detect_firewall_backend` calls `nft list ruleset` with no timeout wrapper. If the nft ruleset is corrupted or nft hangs (e.g., due to a kernel lock), this command blocks indefinitely.

The `rules` ubus call is invoked when the LuCI view loads. If `nft list ruleset` hangs, the entire page rendering blocks — the poll and log display still work (they don't depend on rules), but the rule-name resolution hangs indefinitely.

Compare with `resolve_hostname` two lines later which properly wraps `getent` in `timeout 5`.

## Suggested Fix

Wrap the nft call in `timeout 5`:

```sh
nft_rules=$(timeout 5 nft list ruleset 2>/dev/null) || nft_rules=""
```

**Found by:** Big Pickle in Zen MCR. Confidence: high.

lucas-albers-lz4:
## Summary
- Wrap `nft list ruleset` with `timeout 5` so a hung/corrupted ruleset cannot block LuCI rules load (#61).
- `__rulesmap_iptables` is CLI-only (not a ubus method); restrict it to fixed `/tmp/rulesmap` and update the host test (#54).
- Drop the no-op `sh -c` around `ubus call log read`; call ubus with a fixed JSON blob.
- Escape `wan_zone` / logging blockers via `json_escape` / `json_null_or_string` (#43 leftover).

Stacked on #62.

Closes #61
Closes #54

## Test plan
- [x] `./scripts/fwlive-test.sh`
- [ ] On device: Status → Firewall Live View loads rules map without hang
- [ ] Confirm `ubus call fwlive rules` still returns backend + rules

Made with [Cursor](https://cursor.com)

### DIFF
--- a/openwrt-feed/luci-app-fwlive/root/usr/libexec/fwlive-logging.sh
+++ b/openwrt-feed/luci-app-fwlive/root/usr/libexec/fwlive-logging.sh
@@ -78,10 +78,11 @@ check_nf_log_ipv6() {
 logging_blockers_append() {
 	blocker="$1"
 	[ -n "$blocker" ] || return 0
+	esc=$(printf '%s' "$blocker" | json_escape)
 	if [ -n "$LOGGING_BLOCKERS" ]; then
 		LOGGING_BLOCKERS="${LOGGING_BLOCKERS},"
 	fi
-	LOGGING_BLOCKERS="${LOGGING_BLOCKERS}\"${blocker}\""
+	LOGGING_BLOCKERS="${LOGGING_BLOCKERS}\"${esc}\""
 }
 
 collect_logging_blockers() {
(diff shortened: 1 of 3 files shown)


lucas-albers-lz4:
Superseded by a new PR targeting `master` after this PR auto-closed when #62’s branch was deleted. Continuing the merge session there.

Issue #216, opened 2026-08-31
Umbrella: openwrt/luci#8992 review findings — all 8 valid, +1 missed


lucas-albers-lz4:
## Goal
One umbrella for the review findings on the first `luci-app-fwlive` upstream PR,
[openwrt/luci#8992](https://github.com/openwrt/luci/pull/8992) (commit `91b4629`).

**Verdict: all 8 findings raised by the upstream `openwrt-ai` reviewer are valid.** One additional
defect was found locally that the reviewer missed, and one reviewer nit turns out to be a
blocker-class packaging bug on closer inspection.

Blocks #209 (the luci upstreaming plan). This umbrella is **facts only** — evidence and sizing.
Fixes and implementation plans are separate, per repo convention.

## How this was verified
Every finding was re-checked by **executing the shipped code** at the PR's exact commit under both
target shells (`dash` and BusyBox `ash`), not by reading it. Claims about OpenWrt itself were settled
against upstream `openwrt/openwrt` and `openwrt/luci` sources. Nothing in the upstream review was a
false positive.

## Children

| Issue | Finding | Severity | Effort |
|---|---|---|---|
| #217 | `rules` map lost to pipeline subshell — nft/iptables labels never reach the client | **Blocker** | M · ~2.5–4 h |
| #218 | `getent` absent on OpenWrt — `resolve` permanently dead | **Blocker** | M · ~3–4.5 h |
| #219 | Classifier fork storm — poll cannot finish inside its own 1 s interval | **Blocker** | **L · ~6–12 h** |
| #220 | `jsonfilter` undeclared — silently empty log table on a stock image | **Blocker** | S · ~30–45 min |
| #221 | Poll `lines` clamp bypassable via digit overflow | Medium | S · ~30–45 min |
| #222 | `json_escape` inverted dependency — `prerm` emits `not found` | Low | S · ~30–60 min |
| #223 | Duplicate JSON keys in the `rules` map *(missed upstream)* | Low | S · ~20–30 min |
| #224 | `SOURCE_DATE_EPOCH` no-op + `APP_VERSION` drift | Nit | S · ~45–75 min |
| #225 | Dual-maintenance policy answer + dependency prose | Nit / policy | S · ~30–45 min |
| #226 | UCI rule names with whitespace word-split into junk keys *(found during #217 verification; pre-existing on master)* | Low | S · ~30–45 min |

[...]

PR #258, opened 2026-09-02
fix(ui,css): document warnings reply key; warn-tint the degraded backend span (luci #8992 round 5)

lucas-albers-lz4:
## What

Round-5 follow-up for openwrt/luci#8992 wave-5 review (two client-side nits — no backend change).

1. **`expect` default now documents the full `logging_status` reply shape** — added `warnings: []` alongside `blockers: []` (the key the backend always emits but the client default omitted). `rpc.declare` semantics unchanged: with the `''` key the default only substitutes when the reply is null/non-object.
2. **Degraded backend readouts get warn styling** — `#fwlive-backend` previously rendered rules-map errors and the `timeout_missing` warning at the same 0.65em muted grey as the quiet `using fw4` metadata. `updateBackendUi()` now toggles a `fwlive-backend-warn` class on the existing span when `lastRulesError` or a warning is present (`--warn-color-high`, matching `.fwlive-status-error`); no new element, no layout change.

## Files

- `htdocs/luci-static/resources/fwlive/fwlive.css` — CSS source: `.fwlive-backend.fwlive-backend-warn` rule
- `htdocs/luci-static/resources/fwlive/css.js` — regenerated via `scripts/embed-fwlive-css.js` (freshness-gated)
- `htdocs/luci-static/resources/view/status/fwlive.js` — `warnings: []` in expect; `degraded` flag → `classList.toggle`
- `tests/fwlive-modules-smoke.test.js` — source-level contract assertion for the expect shape

## Verification

Full `fwlive-test.sh` suite passes; theme-CSS freshness test + modules-smoke green.

Refs openwrt/luci#8992, #216

## Summary by CodeRabbit

- **New Features**
  - Backend indicators now show a warning state when diagnostic or rule warnings are present.
  - Detailed map views use a wider layout for improved visibility.
  - Paused states now have dedicated visual styling.
  - Added fallback styling for row tinting and backend warning indicators.
  - Improved message layout behavior for wrapping and single-line display across different screen sizes.


lucas-albers-lz4:
Folded the latest openwrt-ai nit (post-round-5 snapshot review):

- Dropped the inaccurate \`classList.toggle(name, force)\` / 21.02-era browser notes
- Collapsed \`updateBackendUi\` and \`updateMessageLayoutUi\` to the two-arg \`toggle\` API LuCI core already uses

Bugbot: clean. Luna on prior tip asked for this fold (now committed). Tests + baseline green.

PR #264, opened 2026-09-04
fix(lab): require opt-in for insecure SSH in agent-build-and-deploy (#261)

### LINKED ISSUE #261: security: agent-build-and-deploy.sh disables SSH host-key verification (MITM -> malicious ipk as root)
## Summary

`scripts/agent-build-and-deploy.sh` (lab/QEMU deploy helper for `luci-app-fwlive`) disables SSH host-key verification for every `scp`/`ssh` to the target OpenWrt host, so a network-path attacker (ARP/DHCP spoof, wrong-host lease mixup) can silently MITM the connection and swap the uploaded `.ipk`, leading to arbitrary code install as root on the target.

## Evidence

`scripts/agent-build-and-deploy.sh`:

- `:24` — `SSH_OPTS=(-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null)` (hard-coded default, not env-gated)
- `:118-126` — used on every connection: `scp -O ... "${SSH_OPTS[@]}" "$IPK_PATH" ...`, fallback `ssh ... "cat > ${REMOTE_IPK}"`, and `ssh ... "opkg install ${REMOTE_IPK}"` — all as `${OPENWRT_USER}` (default `root`, `:19`)
- `:79-90` — host resolution falls back to lease-file discovery (`discover_ip_from_leases`), so a stale/wrong lease silently targets the wrong machine
- `:10` (usage header) documents the same script against a real device: `OPENWRT_HOST=192.168.1.1 ./scripts/agent-build-and-deploy.sh --ipk path/to.ipk`

## Impact

MITM (or wrong-host from lease mixup) → attacker-controlled `.ipk` installed as root on the OpenWrt guest/device. Blast radius limited: this is a documented lab/QEMU helper, not shipped to devices — but it is also the path a developer would reach for when deploying to a real router on the LAN, where the flag provides no protection against an active on-path attacker. No lab-only warning exists at the point of use.

## Classification

Low severity, facts-only finding. Fix suggestion: keep normal host-key verification by default and require an explicit opt-in (e.g. `ALLOW_INSECURE_SSH=1`) or a `--lab-only` flag with a warning when disabling; QEMU-hostfwd/lease-discovered guests (ephemeral keys) can opt in deliberately.

Found by read-only security review (Muse Spark 1.3 Free / `muse-spark-1.3-contributor-free`, 2026-09-03); verified against `master@1870b71398e454e5896f7248b5362acaaa0a2f50`.

lucas-albers-lz4:
## Summary
- Default SSH host-key verification ON in the lab deploy helper
- Opt in via `ALLOW_INSECURE_SSH=1` or `--lab-only` (warns at use)
- Record control in security-review.md

Closes #261

## Test plan
- [x] `bash -n scripts/agent-build-and-deploy.sh`
- [ ] Manual: without opt-in, known_hosts mismatch fails; with `--lab-only`, warns and proceeds

### DIFF
--- a/docs/developer/security-review.md
+++ b/docs/developer/security-review.md
@@ -1,7 +1,7 @@
 # Security review state
 
-> **Status:** 40 controls in force; 0 open findings.
-> **Last review:** 2026-08-31 (upstream-review remaining).
+> **Status:** 41 controls in force; 0 open findings.
+> **Last review:** 2026-09-03 (lab deploy SSH host-key default).
 > **Open:** none.
 > **Next:** re-check pins before each `v*` tag; close the 4 honest gaps in the lab.
 > **How to verify:** `./scripts/fwlive-test.sh` runs automated host checks. For coverage beyond that script, follow [`.cursor/skills/security-audit/SKILL.md`](../../.cursor/skills/security-audit/SKILL.md). Values current as of this PR.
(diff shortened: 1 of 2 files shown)


lucas-albers-lz4:
@coderabbitai review

Issue #254, opened 2026-09-01
Pointer: regexproof dogfood_shell mining (#615) + fwlive normalize idempotency contract — sequence


lucas-albers-lz4:
# Pointer — regexproof dogfood_shell mining + fwlive idempotency contract

Pointer ticket: the work itself lives in **[regexproof#615](https://github.com/lucas-albers-lz4/regexproof/issues/615)** ("Funnel: mine admitted dogfood_shell corpus + fwlive normalize idempotency contract"). This ticket records what it means for fwlive and the suggested sequence, so the state is visible from this repo.

## Why this matters for fwlive

fwlive's **POSIX-shell regex surface** was already admitted to regexproof's dogfooding universe (gate decision `go`, 2026-08-12 — the first admitted corpus covering the posix-shell dialect), but **never mined**: zero conversion-ledger entries, no materialization, no Smith intake. The corpus includes fwlive's `rpcd/fwlive`, `fwlive-logging.sh`, `fwlive-is-firewall-event.sh`, `fwlive-log-filter.sh`, plus `scripts/` and `tests/` shell files.

The mining is a **dogfooding exercise for regexproof's tooling** (user code is the P(compiles) distribution), not primarily a findings hunt for fwlive — the classifier/address surfaces are already Z3-verified in-tree (F1–F4, #120).

## What the regexproof ticket covers

**A. Mining-state diagnosis (what's blocking Smith):**
1. `corpus_pin: local` — Smith's materialize-at-pin step assumes a public repo + SHA. Dogfood repos are user-owned local paths. Needs a local-path variant or freeze procedure before `batch/manifests.py` can reference them. *Likeliest structural blocker.*
2. Dogfood repos move (fwlive ships releases, re-cuts into openwrt/luci) — a stale freeze invalidates the admission denominator. It needs a pin-refresh rule.
3. `stateful: 329` predicted reject bucket — does the posix-shell backend encode stateful `s///`? The `>= 0.30` encodable fraction gate has never run for this corpus.

**B. The fwlive contract that exercises the conversion hop:**
[...]

PR #280, opened 2026-09-05
test(z3): F5 normalize_log_prefix idempotency property and guard (#275)

### LINKED ISSUE #275: Phase 4: F5 — normalize_log_prefix idempotency in z3-verify.py
# Phase 4: F5 — `normalize_log_prefix` idempotency in `z3-verify.py`

Part of the wave tracked in #271. Source of truth: #254 (Properties, Mutation guards, step 3) and #240 §6 item 5.

## Objective

Encode the idempotency contract in-tree: one property, one mutation guard, seconds-grade, independent of the upstream landing and the regexproof timeline.

## Scope

- Add F5 to `scripts/z3-verify.py`: P1 idempotency + P2 fixpoint on `normalize_log_prefix` (`sed 's/[[:space:]:]*$//'`).
- Add the quantifier-weaken mutation guard (`*` → `?`, replays the exact 2026-08 regression per commit `ed4486829c`). Guard must flip P1/P2 to SAT.
- Label P3 (client-parity data-flow invariant) as out of scope for this phase with a pointer to #254. Do not silently drop it — state where it will land.
- Time `z3-verify.py --full` before and after. It stays in seconds.

## Non-goals

- No regexproof-side encoding (waits for post-#8992 pin stability). No F1–F4 changes. Rides Phase 1, 2, or 3 as a second commit, or ships standalone.

## Acceptance criteria

- [ ] F5 proves P1+P2 on the current function text. `--full` finishes in seconds.
- [ ] Quantifier-weaken guard flips P1/P2 to SAT (negative control proven).
- [ ] P3 recorded as follow-up with an owner, not dropped silently.
- [ ] CI Z3 gate green.

## Depends-on

None. Batches with any of Phases 1–3.

## Effort

S. No upstream gate (function body is path-independent).

lucas-albers-lz4:
Closes #275.

Encodes the idempotency contract in-tree: one property pair, one mutation guard, seconds-grade, independent of the upstream landing and the regexproof timeline.

Changes (scripts/z3-verify.py only, no production code):
- P1 idempotency: strip(strip(s)) == strip(s) over a bounded domain (length <= 8, content + strip alphabet).
- P2 fixpoint: the stripped output never ends in a space, tab, or colon.
- Quantifier-weaken guard: the '?' single-strip model must violate P1/P2 (SAT) — replays the exact 2026-08 regression (ed4486829c), guard witness is a two-trailing-char string in the zz:: shape.
- Shipped-text pin: the proof runs only while rpcd/fwlive still carries sed 's/[[:space:]:]*$//'; any change fails closed with re-verify P1/P2.
- Shell ground truth (--full only): the extracted shipped function runs twice over a 13-input tricky corpus (zz::, tabs, mixed trailing runs); f(f(x)) == f(x) and fixpoint asserted.
- P3 (client-parity data-flow into parseRuleHint) labeled out of scope with a pointer to #254.

Verification:
- --fast green (F5 adds about 4 s). --full green (F5 shell replay instant; total about 61 s vs 56 s baseline, F3-dominated).
- Red-proved: committing '*' -> '?' in the shipped sed fails with F5 shipped sed form changed.
- Rides standalone per the issue (no dependency on phases 1-3). No regexproof-side encoding, no F1-F4 changes.

### DIFF
--- a/scripts/z3-verify.py
+++ b/scripts/z3-verify.py
@@ -11,6 +11,10 @@ boundaries, prefix+boundary, flag-token/ws language, glue site). Not
 ECMA-direct. NETFILTER_KV_GLUE lookahead is string-ops, not the lookahead
 regex. Pinned for stock z3-solver==5.0.0 (default seq backend; no z3str3).
 
+F5: normalize_log_prefix idempotency (P1) + fixpoint (P2) over a bounded
+domain, with a quantifier-weaken guard replaying the 2026-08 regression.
+P3 (client parity) lives in #254, not here.
+
 Usage:
   ./scripts/z3-verify.py --fast
   ./scripts/z3-verify.py --full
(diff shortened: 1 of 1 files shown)


lucas-albers-lz4:
@coderabbitai review

Issue #291, opened 2026-09-06
Umbrella: shipped-surface conventions — SPDX headers + shell strict mode + duplicate-file + Makefile cut + wave/ audit (#288 M3, M5, M6, M8, M11)


lucas-albers-lz4:
## Parent
Cluster from the upstream-readiness review (#288) — M3, M5, M6, M8, M11. Per-file SPDX headers, shell strict-mode audit, the `__duplicate__` file investigation, the Makefile `luci.mk` include invariant, and the `wave/fwlive-190/` snapshot pattern. Each is XS-to-S individually; together they are the "shipped-surface conventions" wave that lands before the upstream PR is filed.

## Goal
Five small conventions/audits that an openwrt/luci maintainer will check first. Combined wall-clock: ~3-4 h. All independent; can land as 1-5 PRs.

## Phase map

| # | Source finding | Scope | Effort |
|---|---|---|---|
| C1 | M3 | Add SPDX-License-Identifier: Apache-2.0 (and a copyright line) to all 12 shipped JS files in `htdocs/luci-static/resources/fwlive/*.js` and the view file `htdocs/luci-static/resources/view/status/fwlive.js` | XS, ~15 min |
| C2 | M3 | Add a 3-line SPDX header check to `fwlive-test.sh` so the next file added without SPDX fails the build | XS, ~10 min |
| C3 | M5 | Audit the rpcd entry point + the 4 libexec shell helpers for `set -e`/`set -u`/`set -o pipefail`; add where safe; document explicit `|| return 1` where strict mode is too strict (acquire_wan_log_lock from #244 is the anchor) | M, 2-3 h |
| C4 | M6 | Run `pygount --format=json` on the shipped surface; confirm the `__duplicate__` file is the intentional `core/` ↔ `htdocs/.../fwlive/log.js` mirror (drift-gated by `gen-luci-wrapper.js`); if anything else, file a follow-up | XS, ~30 min |
| C5 | M8 | Verify Phase 2 of #273 (upstream-cut invariants) lands the Makefile-include rewrite test (`include ../../luci.mk` after cut, not `$(TOPDIR)/feeds/luci/luci.mk`) | XS, verification only |
| C6 | M11 | Audit the `wave/fwlive-190/` snapshot pattern: confirm intentional lab breadcrumb (most likely), add a one-line `wave/README.md` explaining the pattern + a `.gitignore` for build artifacts, OR `git rm -r wave/` if accidental | XS, ~30 min |

[...]

Issue #244, opened 2026-08-31
blocker: acquire_wan_log_lock exec abort + success-path stderr silence


lucas-albers-lz4:
## Parent
Part of umbrella #242 (openwrt/luci#8992 tip `2defca544d`). Blocks #209.
Introduced/exposed on the #238 BusyBox WAN-lock path.

## Fact
`acquire_wan_log_lock` ends with:

```sh
exec 9>"$WAN_LOG_LOCK_FILE" 2>/dev/null || return 1
```

Three defects, all verified under `dash` and BusyBox `ash`:

1. **`|| return 1` never runs on open failure.** A redirection error on `exec`
   with no command aborts a POSIX non-interactive shell. Under rpcd that kills
   the plugin mid-call (no `{"ok":false,"error":"lock_failed"}` body); in
   `prerm` it can abort before the fail-open `exit 0`.
2. **`2>/dev/null` does not hide the diagnostic** — redirections apply left to
   right, so `9>` is attempted before stderr is reassigned.
3. **On the success path `2>/dev/null` sticks.** Later `uci` / `flock`
   diagnostics in this process are discarded for the rest of the lifetime;
   `release_wan_log_lock` does not restore stderr.

## Evidence

```
$ dash -c 'acq(){ exec 9>"/nonexistent-dir/x.lock" 2>/dev/null || return 1; }; \
    if acq; then echo acquired; else echo "acq failed"; fi; echo "still alive"'
dash: 1: cannot create /nonexistent-dir/x.lock: Directory nonexistent
# "still alive" never prints; outer exit 2

$ busybox sh -c '...same...'
sh: can't create /nonexistent-dir/x.lock: nonexistent directory
# "still alive" never prints; outer exit 1
```

Success-path stderr silence: after a successful `exec 9>lock 2>/dev/null`, a
later `echo test >&2` produces no stderr output.

Reviewer-suggested probe works:

```sh
( exec 9>>"$WAN_LOG_LOCK_FILE" ) 2>/dev/null || return 1
exec 9>>"$WAN_LOG_LOCK_FILE"
```

Fail path returns 1 and keeps the shell alive; success path keeps stderr.

## Impact
WAN enable/disable can return an empty rpcd body instead of a structured error;
uninstall restore can abort early; successful lock holders lose stderr for the
rest of the call. Contract comment ("fails closed only if lock cannot be opened")
is false today.

## Size
**S** — subshell probe + `>>` (or equivalent); host test for fail path that
[...]

PR #296, opened 2026-09-06
fix: rpcd/libexec strict mode audit (#291 C3)

lucas-albers-lz4:
## Summary
- Enable `set -eu` on the rpcd plugin and log-filter entry points, with a subshell probe for `pipefail` (BusyBox ash has it; Debian dash used as host `sh` rejects `set -o pipefail` as a startup abort — same class as the #244 failed-`exec` lesson).
- Leave sourced helpers (`fwlive-logging.sh`, generated `fwlive-is-firewall-event.sh`) without blanket `set -euo`: prerm sources logging.sh alone and is best-effort (`restore … || logger`; exit 0).
- Document explicit `|| return 1` / `|| true` sites where strict mode is too strict (uci miss, unset WAN log option, NXDOMAIN, jsonfilter miss that must still close `{"log":[…]}`).
- `collect_logging_blockers` / `collect_logging_warnings` now report via globals and always return 0 — a `return 1` “has blockers” status aborted `build_logging_status_json` under dash/ash `set -e`.

Refs #291

## Test plan
- [x] `bash scripts/fwlive-shellcheck.sh`
- [x] `bash tests/fwlive-logging.test.sh`
- [x] `sh` and `busybox ash` `rpcd/fwlive __selftest` (includes new status-JSON smoke under set -e)
- [x] `bash tests/fwlive-logging-lock.test.sh` (#244 lock/exec paths)
- [x] `node tests/fwlive-rules-map.test.js`, `fwlive-shell-filter.test.js`, `fwlive-codegen.test.js`
- [x] `./scripts/fwlive-test.sh` (full suite, prior apply of the same diff)

## Residual risks
- Device-only `jshn` paths (`poll_lines_from_input` / `resolve_addresses`) are guarded with `${first:-}` / `${ip:-}` but not exercised on host CI.
- `validate-openwrt.sh` on-device check is the remaining verification for ash + real ubus/uci.

### DIFF
--- a/openwrt-feed/luci-app-fwlive/root/usr/libexec/fwlive-is-firewall-event.sh
+++ b/openwrt-feed/luci-app-fwlive/root/usr/libexec/fwlive-is-firewall-event.sh
@@ -4,6 +4,7 @@
 # GENERATED FILE — do not edit. Run: ./scripts/gen-all.sh
 # source: core/fwlive-log.js CLASSIFY_SPEC
 # Shared isFirewallEvent parity logic (shell). Sourced by fwlive-log-filter.sh and tests.
+# Sourced library: do not add set -euo here (callers own strict mode, #291 C3).
 # One awk process classifies a batch (MODE=json) or one message (default).
 
 _fwlive_run_classify() {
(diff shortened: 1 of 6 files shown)


lucas-albers-lz4:
Luna re-review: APPROVE after SC2015 + find_wan_zone fixes. Marking ready.

Issue #303, opened 2026-09-06
test: rpcd fail-closed structured error contract (host selftest)


lucas-albers-lz4:
## Parent

Confirmed **do** from [#302](https://github.com/lucas-albers-lz4/fwlive/issues/302) (T2-A). Source: #288 Tier-2 leftover. Luna + Grok both **DO** (high).

## Goal

Host-side contract: every rpcd method’s fail-closed path emits a structured `error` field (not a silent empty success shape). Prevents the silent-failure / empty-reply class from the bug review (`#220`, `#233`, `#245` and kin).

## Scope

- Prefer **host tests** (extend `tests/fwlive-rpcd-security.test.js` or a sibling) that drive production `rpcd`/`__selftest` paths and assert `error` presence/shape on forced failures.
- Do **not** grow shipped `__selftest` bytes for the luci cut tree unless a tiny existing hook already covers a method — the monorepo CI is the gate; upstream luci will not run our host suite.

## Acceptance

1. Inventory of rpcd methods and which fail-closed paths currently omit `error` (document in the PR).
2. Host tests pin: on induced failure, reply JSON includes non-empty `error` (and does not look like a successful empty payload alone).
3. Green under `./scripts/fwlive-test.sh` (or the existing rpcd test entry).
4. No new shipped surface required for upstream cut size.

## Non-goals

- semgrep / shellharden / spellcheck (explicit **don't** on #302).
- Changing happy-path reply shapes unless required to attach `error` on failure.
- Expanding device-shipped `__selftest` as the primary proof vehicle.

## Effort

M (~2–3 h) per #288 estimate.

## Provenance

- #288 Tier-2 `fwlive-rpcd-selftest` extension
- #302 T2-A confirmed 2026-09-06 (Luna+Grok agree DO; maintainer confirmed)

---

## Review fold-in — astra findings adjudicated (2026-09-09)

Multi-model review (zen MCR) + ground-truth verification confirmed the review findings; the following is folded into the issue contract. Documentation-only; PR #305 code stands.

1. **Contract exceptions (part of the contract, not violations).** `logging_status` intentionally reports readiness via `ready`/`blockers`/`warnings` without an `error` field (pinned by `testLoggingStatusNeverSilent`). Normal NXDOMAIN/absent PTR remains distinguishable from infrastructure failure: the name is simply omitted from `names`; `error` is emitted only for `no_resolver` / `jshn_missing` / `invalid_input`.
[...]

Issue #245, opened 2026-08-31
bug: rules map error field never reaches the UI


lucas-albers-lz4:
## Parent
Part of umbrella #242 (openwrt/luci#8992 tip `2defca544d`). Related to #231 / #238 bounds work.

## Fact
`build_rules_map` can emit `error` (`mktemp_failed`, `rules_truncated` once
`RULES_MAP_MAX_KEYS` / `RULES_MAP_MAX_BYTES` trip). `loadRulesMap` only reads
`res.rules` / `res.backend`:

```js
this.rulesMap = (res && res.rules) || {};
this.firewallBackend = (res && res.backend) || 'nft';
```

A truncated or empty map renders identically to a router with no labelled rules.
Same commit dropped `expect` on poll specifically so `reply.error` could reach
the banner (#233); rules did not get the matching client treatment.

## Evidence
Code path read on tip `a30a884` / luci `2defca544d`. Truncation is reachable:
`map_uci_rule_names` adds up to two keys per UCI rule before nft prefixes, so
~256 firewall rules exhaust the 512-key budget and later nft `log prefix`
labels are dropped with no UI signal.

## Impact
Silent partial map — operators cannot tell labels are missing because of
truncation vs. absent. Nit per reviewer; Medium in practice once bounds exist.

## Size
**S** — surface `res.error` like poll (`lastPollError` / banner / console). Est. 45–75 min.

## Acceptance
- [ ] `rules_truncated` / `mktemp_failed` visible in UI or at least logged
- [ ] Happy path unchanged
- [ ] `./scripts/fwlive-test.sh` green

lucas-albers-lz4:
Fixed on master in 39e83a8 (#247); luci recut 593ce11d20.

Issue #233, opened 2026-08-31
bug: poll error field never reaches the UI (rpc.declare expect log:[])


lucas-albers-lz4:
## Parent
Opus final review of #228. Not blocking the luci cut: `jsonfilter` is now a hard `LUCI_DEPENDS`, so the #220 empty-table case is fixed at packaging. Umbrella #216. Related closed work: #55 (poll exceptions swallowed).

## Fact
`callFwlivePoll` is declared with `expect: { log: [] }`. LuCI’s rpc helper therefore returns only the `log` array. `fetchEntries` sets `lastPollError` on throw or a non-array; a well-formed `{"log":[],"error":"filter_failed"}` (or `jsonfilter_missing`) is treated as a successful empty poll and **clears** `lastPollError`.

The error still exists on the ubus wire and in syslog (`logger -t fwlive`). The table just goes empty with no banner. Same shape as #220 wanted to stop, for every *other* filter failure (awk crash, `jsonfilter -s` ARG_MAX, future `error` values).

`resolve` already surfaces `error:no_resolver` in JSON; `poll` has the same field and the client drops it.

## Size
**S** — stop using `expect: { log: [] }`, or read the full object and set `lastPollError` / a banner from `error`. Est. 30–60 min.

## Acceptance
- [ ] `{"log":[],"error":"filter_failed"}` shows the poll-error banner (or equivalent), not a silently empty table
- [ ] Happy-path `{"log":[…]}` still updates the table
- [ ] Host or E() harness coverage for the error object
- [ ] `./scripts/fwlive-test.sh` green

## Non-goals
Changing the rpcd `error` strings themselves. #231 item 1 (rules-map mktemp-skip `error` field) is a sibling, not this UI drop.

Issue #220, opened 2026-08-31
blocker: jsonfilter is an undeclared dependency — log table silently empty on a stock image


lucas-albers-lz4:
## Parent
Part of umbrella #216. Blocks #209.

## Fact
`fwlive-log-filter.sh` requires `jsonfilter`:

```sh
if ! command -v jsonfilter >/dev/null 2>&1; then
	printf '%s' '{"log":[]}'
	exit 0
fi
jsonfilter -s "$input" -e '@.log[*]' ...
```

`jsonfilter` is **not declared** anywhere: `LUCI_DEPENDS:=+luci-base +logd`, and the README Dependencies
section lists only `luci-base` and `logd`.

Verified that it is **not guaranteed present** on a stock image:

- **Not** in `DEFAULT_PACKAGES` (`openwrt/openwrt` `include/target.mk`): base-files, ca-bundle, dropbear,
  fstools, libc, libgcc, libustream-mbedtls, logd, mtd, netifd, uci, uclient-fetch, urandom-seed, urngd.
- **Not** pulled by `luci-base` — its `LUCI_DEPENDS` is rpcd, rpcd-mod-file, rpcd-mod-luci, rpcd-mod-ucode,
  cgi-io, ucode + ucode mods, liblucihttp-ucode. No `jsonfilter`.
- **Not** pulled by `netifd` — `DEPENDS:=+libuci +libnl-tiny +libubus +ubus +ubusd +jshn +libubox
  +libudebug +ucode +ucode-mod-fs +ucode-mod-ubus +ucode-mod-uloop +ucode-mod-uci`. No `jsonfilter`.
- `jsonfilter` is its own Base-system package (`package/utils/jsonfilter`, `DEPENDS:=+libubox +libjson-c`).
- **Precedent in the same feed:** `applications/luci-app-pbr/Makefile` declares
  `LUCI_DEPENDS:=+luci-base +jsonfilter +pbr`.

## Impact
On an image without `jsonfilter`, the guard makes the filter return `{"log":[]}` and **exit 0**. The log
table is silently empty with no error surfaced anywhere in the UI — the app appears installed and working
but shows nothing, forever. Same silent-failure class as the `getent` finding (child B).

[...]

PR #305, opened 2026-09-07
fix(rpcd): fail-closed structured error on every method (#303)

lucas-albers-lz4:
Add a structured `error` field to every rpcd method's fail-closed path so failures do not look like silent empty successes.

## What changes

`openwrt-feed/luci-app-fwlive/root/usr/libexec/rpcd/fwlive`:
- `rules` — `nft_failed` / `iptables_failed` / `ip6tables_failed` on dump failure, `no_backend` when no firewall backend is usable. IPv4 enrichment is preserved when only the IPv6 dump failed (partial success stays partial, not silent).
- `poll` — `log_read_failed` on dead logd (bypasses the filter instead of shaping an empty table), `filter_empty` on empty filter output.
- `resolve` — `jshn_missing` / `invalid_input` instead of bare `names:{}`.
- `logging_status` — unchanged: state travels via `blockers` / `warnings` with `ready:false` by design.
- Toggles — `no_wan_zone` already returned the error; the new host test pins the pre-lock short-circuit.

## Tests

`tests/fwlive-rpcd-security.test.js` adds host contract tests that drive the production `rpcd` script with stubbed PATH entries to force each fail-closed path. Coverage:

- `testUnknownMethod` — unknown method exits non-zero with `error`.
- `testRulesNoBackend` — no firewall backend reachable surfaces `no_backend`.
- `testRulesNftDumpFailure` — stateful nft stub: detect probe succeeds, dump fails → `nft_failed`.
- `testRulesIptablesDumpFailure` — IPv4 dump fails → `iptables_failed`.
- `testRulesIp6tablesDumpFailure` — IPv4 succeeds, IPv6 fails → `ip6tables_failed` AND IPv4 keys preserved.
- `testPollUbusFailure` — dead logd stub → `log_read_failed` with the `log` shape intact.
- `testResolveJshnMissing` — isolated PATH → `jshn_missing`.
- `testResolveInvalidInput` — only runs where `jshn` exists; otherwise skips.
- `testLoggingStatusNeverSilent` — pins the full 8-key shape so a regression can never degrade into a silent empty object.
- `testToggleNoWanZone` — pins `FWLIVE_WAN_LOG_LOCK_FILE` and asserts the lock file is untouched on the no-zone short-circuit.

No `__selftest` growth; no new shipped surface for the upstream cut size.

## Out of scope

UI consumption of the new `resolve` errors. `callFwliveResolve` uses `expect: { names: {} }` which drops sibling fields; the new fail-closed codes are an rpcd contract, not a UI one. The existing `no_resolver` pre-check still reaches the view. Follow-up to surface the new codes in the LuCI view can be a separate change.

`logging_status` deliberately has no `error` field — its design is full-shape plus `blockers` / `warnings` with `ready:false`. A regression that drops a key now fails the test.

## Acceptance

Issue AC1: inventory + tests cover the methods and fail-closed paths the bug review surfaced (`#220`, `#233`, `#245` and kin).
Issue AC2: host tests assert `error` presence and non-empty shape on every induced failure; the test for the silent-empty class (`log_read_failed`) explicitly checks the reply does not look like a successful empty payload.
Issue AC3: `./scripts/fwlive-test.sh` is green.
Issue AC4: no new shipped surface for the upstream cut size.

Refs lucas-albers-lz4/fwlive#303

## Summary by CodeRabbit

* **Bug Fixes**
  * Firewall and logging failures now return clear, structured error messages instead of silently producing incomplete results.
  * Missing backends, invalid input, unavailable tools, empty filter results, and missing WAN zones are reported explicitly.
  * Unknown RPC methods now return structured errors.
  * Logging status responses consistently include readiness, blocker, and warning details.

* **Documentation**
  * Added documentation covering fail-closed error behavior and supported error responses.


lucas-albers-lz4:
@coderabbitai review

lucas-albers-lz4:
@coderabbitai review

lucas-albers-lz4:
@coderabbitai review

Issue #243, opened 2026-08-31
blocker: resolve client double-unwraps names — Show hostnames still dead


lucas-albers-lz4:
## Parent
Part of umbrella #242 (openwrt/luci#8992 tip `2defca544d`). Blocks #209.

## Fact
`callFwliveResolve` declares `expect: { names: {} }`. luci-base `rpc.js` unwraps
that before resolving the promise (`ret = ret[key]`, then type fallback to the
default). So the awaited `res` **is** the names map (`{"8.8.8.8":"dns.google"}`).

The client then does:

```js
const names = (res && res.names) || {};
```

`res.names` is always `undefined`, so `names` is always `{}`. Every IP takes the
`hostname.failMark` path; the table never shows a PTR name.

## Evidence
Simulated against the same unwrap rules as luci-base `rpc.js`:

```
after expect unwrap: {"8.8.8.8":"dns.google"}
res.names: undefined
client names = (res && res.names) || {} => {}
fixed names = res || {} => {"8.8.8.8":"dns.google"}
```

Also: with `expect` kept, an error reply `{"names":{},"error":"no_resolver"}`
collapses to `{}` and the `error` field is discarded (same pattern #233 fixed
for poll by dropping `expect`).

## Impact
The `getent` → BusyBox `nslookup` backend fix (#218 / #228) works, but the
advertised "Show hostnames" feature is still dead end-to-end. Reviewer-rated
hard blocker; confirmed.

## Size
**S** — either `const names = res || {}` (keep expect) or drop `expect` and keep
`res.names` while handling `res.error`. Est. 30–60 min including a unit/test
or comment that locks the unwrap contract.

## Acceptance
- [ ] With a working resolver reply, hostnames populate from the names map
- [ ] Missing resolver / empty map still fail-marks without throwing
- [ ] Contract matches luci-base `rpc.js` expect unwrap (no double unwrap)
- [ ] `./scripts/fwlive-test.sh` green

lucas-albers-lz4:
Fixed on master in 39e83a8 (#247); luci recut 593ce11d20.

PR #329, opened 2026-09-12
feat(#306): Layer 1 adaptive poll shedding (always-on)

lucas-albers-lz4:
## Summary
- Always-on server-side adaptive poll cap from measured processing duration (cold/cool/warm/hot); no UCI / no on-device product config
- Test/triage disable only: `FWLIVE_ADAPTIVE=0|false|off|no` or sentinel `/tmp/fwlive-adaptive-off`
- Sibling `.lock` + atomic state rename; hot `resolve` returns `disabled:load`; drop dead `oneshot` from production `log.read`

## Test plan
- [x] `./scripts/fwlive-test.sh` (host suite green)
- [x] `tests/fwlive-adaptive-cap.test.sh` (buckets, duration≠inter-arrival, sibling lock, concurrent writers)
- [ ] QEMU flood AC on #306 (raised `log_buffer_size`, adaptive on vs off) — still open on the parent issue
- [ ] Luna + Bugbot on branch (done pre-PR; re-check after push if needed)

Refs #306

## Summary by CodeRabbit

- **New Features**
  - Added adaptive polling that adjusts firewall log retrieval limits based on processing load.
  - Poll responses now report adaptive status, shed or truncated results, and message counts.
  - Address resolution pauses temporarily after high-load polling to help protect system responsiveness.
  - Adaptive state persists safely across polls and recovers gracefully from unavailable or invalid state data.

- **Documentation**
  - Updated the security review ledger with the new adaptive polling controls.

- **Tests**
  - Added coverage for adaptive limits, load handling, persistence, locking, and failure responses.


lucas-albers-lz4:
## Hermes review pass — Layer 1 adaptive poll cap (Layer 1 of #306)

Reviewed `8d3b07b33c` against `master@9c44c5061c`. CI green; unit suite passes locally (15/15 checks in `tests/fwlive-adaptive-cap.test.sh`). **No blocking findings.** The two questions and three minors below are for the dev to answer/resolve before marking Ready and triggering CodeRabbit.

### ✅ Verified independently (no action needed)
- `oneshot` is genuinely dead on the non-stream `log read` path — the flag is consumed only in logd's stream branch. The grep-gate test pins it.
- Sibling `.lock` design is correct: atomic `mv` replaces the state inode, so locking the JSON inode could not serialize writers. Tests pin lock-inode stability across state rename and exclusivity during a hold.
- Fail-open discipline holds: symlink refusal on state and lock paths, dir-exists + dir-owner checks, `umask 077` + atomic rename, digit/lowercase-alpha sanitizing on every parsed field, missing/busy `flock` → unlocked. No `eval`/`source` of state content.
- Hot path is ash builtins only (no `sed`/`cat`/`jsonfilter`); one `flock` exec per update; bucket boundaries in tests (99/100, 199/200, 800/801) match the constants.
- Reply-shape compatibility: `callFwlivePoll` has no `expect:` strip, so new fields ride through and the view ignores them. `resolve`'s `disabled:load` is eaten by `expect: { names: {} }` — same documented rpcd-vs-UI contract split as the #243 `error` case, and it is called out in the security-review control table.
- Local host-suite failures seen during this review (`jshn-compat` needs real jshn installed; a dash-vs-bash quirk in the rpcd security harness; fork-census poll=7≠6 without jshn) reproduce identically on master — pre-existing VM gaps, not introduced by this PR. CI covers them properly.

### ❓ Questions (dev to answer)
- [ ] **Q1 — A failed `ubus log read` records ~0 ms and reclassifies to `cold`.** If logd is alive but slow (hot state recorded), then fails fast, the next `plan` sees cold state: the cap lifts to full request and `resolve` un-sheds immediately. An oscillating logd (slow → fail → slow) repeatedly sheds and re-lifts its cap, defeating the point of Layer 1 during exactly the incident it exists for. If "failure ≠ health, and a dead logd makes shedding moot" is the deliberate call, say so in a file comment; otherwise consider recording failed reads as neutral (keep previous bucket/limit) instead of a 0 ms success.
- [ ] **Q2 — Nothing in the UI surfaces the new server-side state.** `adaptive`, `shed`, `truncated`, `messages_received`, and `disabled:load` are consumed by no shipped code; the flood banner remains the client-side throttle only. Users under an active server cap will just see a shorter table with no explanation. Is surfacing shed state (e.g. "showing 250 of N requested") a tracked later layer in #306? If yes, link the follow-up here so the AC list distinguishes "not yet" from "overlooked."

### 💡 Minors (fix or record a decision)
- [ ] **M1 — Sentinel is world-creatable.** `/tmp/fwlive-adaptive-off` can be touched by any local user to switch shedding off. Impact is performance-only (no auth/security weakening), so low severity — but a root-only path (e.g. next to the state file under `/var/run`, same owner checks) keeps the disable knob single-tenant. Either move it or note the accepted tradeoff in the header comment.
- [ ] **M2 — Lock-busy fail-open double-writes are accepted but not pinned.** `fwlive_adaptive_with_lock` runs the body unlocked when `flock -n` fails. Under concurrent writers that means two unlocked read-modify-writes can interleave (last-writer-wins on a corrupt-free line, so state stays valid — the convergence test covers the shape, not the ordering). Fine as a budget decision; worth one sentence in the comment saying "busy ⇒ last-writer-wins is acceptable."
- [ ] **M3 — `plan` uses positional `set --` on unquoted command substitution.** Safe today (all six fields are validated numeric/alpha and the bucket words contain no glob chars), but a future field with spaces or a glob metachar in `bucket` would word-split/glob-expand silently. Consider `IFS=' ' read -r` into named vars, or a comment pinning the no-globs invariant. Non-blocking.

### Test plan status (from PR body, for the merge gate)
- [ ] QEMU flood A/C on #306 (raised `log_buffer_size`, adaptive on vs off) — still the open device-pass AC.
- [ ] CodeRabbit round after marking Ready (draft ⇒ auto-review skipped; current round shows "Draft PR not reviewed").

---
*Reviewed by Hermes Agent. Refs #306.*

lucas-albers-lz4:
## Hermes triage

| ID | Disposition | Action |
|---|---|---|
| **Q1** failed read → cold | **CONFIRMED** | Fixed: `log_read_failed` no longer calls `fwlive_adaptive_record` — prior hot/shed persists. |
| **Q2** UI does not surface shed | **DISMISS (this PR)** | Layer 1 is server-only by design. Client surfacing / flood banner is Layer 2 on #306 (not overlooked). |
| **M1** world-creatable `/tmp` sentinel | **FOLD** | Default sentinel is now `<state-dir>/fwlive-adaptive-off` (production: `/var/run/fwlive-adaptive-off`). |
| **M2** lock-busy last-writer-wins | **FOLD** | Documented in `fwlive-adaptive-cap.sh` header + `with_lock` comment. |
| **M3** `set -- $(…)` | **DISMISS** | Fields already sanitized numeric/alpha; no change. |

Pushed on this branch. Still draft until human Ready → CodeRabbit. QEMU flood AC remains on #306.

lucas-albers-lz4:
## CodeRabbit triage

| ID | Disposition | Action |
|---|---|---|
| **CR1** lock open fail skips body | **CONFIRMED** | Fixed: probe `: >>lock` before subshell; on failure run `"$@"` unlocked. |
| **CR2** e2e hot→failed poll→resolve | **CONFIRMED** | Added `testAdaptiveHotSurvivesFailedPoll` in `fwlive-rpcd-security.test.js`. |
| **CR3** parent env forces `adaptive:0` | **CONFIRMED** | Pin `FWLIVE_ADAPTIVE=1` + absent `OFF_FILE` in jshn-compat + poll failure test. |

Pushed.

lucas-albers-lz4:
## Grok → Luna gate

Grok **REQUEST_CHANGES** (P1 post-filter `count_log` scan). Fixed in `6b9c81a`:
- Layer 1 `messages_received:0` (no ash scan after duration clock)
- Lock created `0600`; skip `record` on uptime wrap
- Fork-budget residual deferred (not claimed fixed)

Luna re-review: **APPROVE** — merge OK with fork-budget residual deferred.

Issue #240, opened 2026-08-31
Test plan: practical coverage after upstream (unit vs Playwright balance)


lucas-albers-lz4:
# fwlive practical test coverage plan

> Goal: add tests where they have the best return, not an arbitrary percentage.
> Balance: unit and CLI tests run in seconds. Playwright and QEMU tests run in minutes and can flake. Nothing here raises the per-PR gate beyond the current host suite unless the return justifies it.

### Inventory

| Layer | LOC | Tests today | Runtime |
|---|---|---|---|
| `core/fwlive-log.js` + `resources/fwlive/log.js` (classification, normalize, filter) | ~1,036 | Z3 proofs (F1–F4) + JS unit + BusyBox parity + codegen freshness + shell-filter parity | <1 s |
| `resources/fwlive/*.js` modules (buffer, hostname, chips, table, logging, proto, tint, constants, links) | ~1,120 | modules-smoke + targeted unit (pause/resume, hostname-cache, proto-filter, tint, i18n, theme-css, e-harness) | <1 s each |
| `view/status/fwlive.js` (poll, render, filters, hostname resolve, flood throttle) | 1,764 | `fwlive-view-poll-error.test.js` + `fwlive-view-poll-guard.test.js` + `fwlive-view-smoke.mjs` (mocked Playwright) | ~15 s with browser |
| `rpcd/fwlive` + `fwlive-logging.sh` + `fwlive-log-filter.sh` | 1,759 | shellcheck + 4 shell test files (logging, logging-lock, rules-map, firewall-filter) | ~2 s |
| Build/publish/supply-chain (docker-sdk, feed signing, digests) | ~8 scripts | feed-keys-mode, fetch-pin-gate, sdk-matrix-digests, feed-publish-assets | ~2 s |
| QEMU/Playwright integration (3 smoke `.mjs` + 5 `qemu-*.sh` wrappers) | — | `chip-invert`, `proto-ui`, `ui-reliability` | 3–8 min per cell, boot required |

The shell boundary breaks down as follows: `rpcd/fwlive` is 902 lines, `fwlive-logging.sh` is 827, `fwlive-log-filter.sh` is 30. Total 1,759 lines.

[...]

Issue #273, opened 2026-09-05
Phase 2: Tier-1 Gaps 4+5 — upstream-cut invariants + .pot parity


lucas-albers-lz4:
# Phase 2: Tier-1 Gaps 4+5 — upstream-cut invariants + `.pot` parity

Part of the wave tracked in #271. Source of truth: #240 §3 Tier 1 (Gaps 4, 5) and the round-2 synthesis.

## Objective

Give the re-cut pipeline and the translation template the same seconds-grade pinning the shell boundary already has, with measured wall time.

## Scope

- Gap 4: run `upstream-cut.sh` to a temp dir and assert the luci-shaped invariants: Makefile include rewritten to `../../luci.mk`, no `SOURCE_DATE_EPOCH` block, no docker-sdk references, README link rewritten to the blob URL, `po/templates` present, locale dirs dropped, `.pot` reference paths `applications/`-prefixed.
- Gap 5: run `i18n-scan.pl` on a cut tree and assert msgids match the checked-in `.pot` exactly (paths normalized).
- Measure wall time on `ubuntu-latest` before merge. The OpenWrt-tree scanner availability is unbudgeted — confirm it runs in CI or scope the test to what CI provides.

## Non-goals

- No Gap 2/3 work (Phase 1). No cut-script redesign — pin current output shape only.

## Acceptance criteria

- [ ] Temp-dir cut asserts all seven luci-shaped invariants. Failure names the invariant.
- [ ] `.pot` msgid parity passes with paths normalized. A changed UI string without `.pot` update fails the test.
- [ ] Measured wall time recorded in the PR. Per-PR gate stays seconds-grade.
- [ ] `fwlive-test.sh` green. Shellcheck green.
- [ ] Rebased on #268 after it merges (keep both test entries on conflict).

## Depends-on

None. Parallel-safe with Phases 1 and 3.

## Effort

M. Gated on openwrt/luci#8992 landing (parked until upstream lands, per plan).

lucas-albers-lz4:
Phase 2 implemented in PR #278 (branch fix/273-phase2-cut-pot, tests only: new tests/fwlive-upstream-cut.test.sh + fwlive-test.sh wiring).

[...]

Issue #239, opened 2026-08-31
WAN logging: @zone[N] vs cfgXXXX in uci changes causes false firewall_changes_pending


lucas-albers-lz4:
## Summary

On stock OpenWrt images with anonymous firewall zone sections, `find_wan_zone_section()` returns `@zone[N]` (e.g. `@zone[0]`), but `uci changes firewall` often reports staged lines using the resolved config id (e.g. `firewall.cfg03dc81.log='1'`).

`wan_log_foreign_staged_lines()` in `fwlive-logging.sh` treats only `firewall.${zone}.log=` as "ours". When `zone` is `@zone[0]` and the staged line is `firewall.cfg03dc81.log='1'`, our own change looks foreign → enable/disable aborts with `firewall_changes_pending`.

## Impact

- Enable / Disable WAN logging dead on anonymous-zone images
- Screenshot capture needed a UCI fallback workaround (`scripts/capture-fwlive-screenshots.mjs`)

## Repro (host test)

`tests/fwlive-logging.test.sh` — section **"@zone[N] vs cfgXXXX in uci changes"** (~line 586):

- Stub `find_wan_zone_section` path via `@zone[0]`
- Stub `uci changes` to emit `firewall.cfg03dc81.log='1'`
- Expect `enable_wan_logging` → `{"error":"firewall_changes_pending"}` and no commit

Currently passes as a **repro** (documents bug). After fix, test should expect `"ok":true` and commit.

## Likely fix

Resolve `@zone[N]` to cfg id (or normalize both sides) before comparing staged lines in `wan_log_foreign_staged_lines` / `wan_log_count_our_staged_lines`.

## Related

- PR #238 (adds repro test, not the fix)
- #209 upstream tracking

Issue #232, opened 2026-08-31
bug: wan_log_lock_dir_safe uses stat -c, which default OpenWrt BusyBox omits


lucas-albers-lz4:
## Parent
Follow-up from the grok last-look of #227 (`stat -c` sticky check). That site was folded to POSIX `[ -k ]`. The same BusyBox `STAT=n` miss remains on the WAN-log lock path. Not blocking #227/#228. Umbrella #216.

## Fact
`wan_log_lock_dir_safe` in `fwlive-logging.sh` requires GNU/BusyBox `stat -c '%u'` and `stat -c '%a'` to prove `/etc/fwlive` is root-owned with no group/other write (#204). Default OpenWrt BusyBox has `STAT=n` / `FEATURE_STAT_FORMAT=n` (same evidence as the #227 sticky fold). `stat` then fails, the helper returns 1, and `acquire_wan_log_lock` fail-closes unless `FWLIVE_WAN_LOG_LOCK_FILE` is set (tests only).

Fail-closed is safe (no lock, no UCI write), but **enable/disable WAN logging is dead on a stock image**. Host CI has GNU `stat`, so the logging-lock tests cannot see this.

POSIX `[ -O ]` covers the uid-0 check when rpcd runs as root. There is no portable `test` for “group/other write bits off”; that part needs a BusyBox-safe substitute (`find -prune -perm`, or drop the mode probe if `[ -O ]` + `mkdir` umask 077 is enough).

## Size
**S** — replace `stat -c` with builtins / `find`; add a PATH-without-stat host case. Est. 45–75 min.

## Acceptance
- [ ] `wan_log_lock_dir_safe` / `acquire_wan_log_lock` succeed on a BusyBox with no `stat` applet when `/etc/fwlive` is a real root-owned directory
- [ ] Symlink / non-root / group-or-other-writable dir still fail closed
- [ ] Host test covers `PATH` with `stat` shadowed
- [ ] `./scripts/fwlive-test.sh` green

## Non-goals
Re-opening the accepted residual create-then-`>` TOCTOU on rules-map dumps (ledger R4). The `_fwlive_tmp_dir_ok` `[ -k ]` fold already landed on #227.

PR #333, opened 2026-09-13
docs: establish defect-driven test approach

lucas-albers-lz4:
## Summary

Document the project’s defect-driven, risk-based test approach based on the criteria developed in #240.

The document preserves the existing practical decisions:

- no percentage coverage target;
- choose the least costly environment that preserves the needed production semantics;
- host tests for the normal per-PR path;
- focused mocked-browser coverage for DOM/LuCI-style behavior;
- focused QEMU/OpenWrt checks for installed-system seams, release, and upstream sign-off;
- no routine per-PR QEMU or broad browser matrix.

It adds the criteria needed to apply those decisions consistently:

- distinguish implemented, executed, enforced, and manually verified evidence;
- require a regression check for confirmed defects unless the ledger records why not;
- escalate only for a distinct boundary or false confidence;
- consider consequence, history, fidelity, detectability, and maintenance cost;
- cover failure/recovery and bounded-work behavior;
- record defer/revisit decisions in a lightweight coverage ledger.

The document also includes calibration examples for POT/source parity, jshn/BusyBox compatibility, and Layer 2 polling races.

Refs #240

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive testing approach guide covering defect-driven test selection, evidence states, test fidelity levels, execution tiers, and coverage reviews.
  * Linked the testing approach guide from developer, build-and-test, and contributing documentation.
  * Clarified testing criteria, environment selection, ownership of guidance, and completion expectations for coverage reviews.


lucas-albers-lz4:
Astra completed an independent strategy review of #240. The draft keeps its core judgment intact and tightens four areas: evidence must distinguish implemented/executed/enforced/manual; harness fidelity must be explicit; escalation requires a distinct boundary or false confidence; and every deferral needs a reason and revisit trigger.

Please review the criteria itself, especially the three calibration cases in the document:

- POT/source freshness;
- jshn/BusyBox compatibility;
- Layer 2 polling races and recovery.

Those examples are intended to expose ambiguity before the later project-wide coverage audit.

lucas-albers-lz4:
## Second-pass review (criteria + calibration)

Astra already tightened the four policy gaps (evidence states, harness fidelity, escalation, deferral). This pass treats the document as durable policy and asks whether the three calibration cases actually force a **single** outcome before a project-wide audit. They currently name the traps well, but they do not fill the ledger, so two reviewers can still pick different dispositions.

The core rule, evidence states, and “least costly environment that preserves the failing semantics” are sound. I would keep those. The notes below are for additional consideration, not a request to expand this PR into the audit itself.

### Calibration cases

**1. POT / source freshness**

The warning is correct: a msgid catalog match is not freshness if the extractor is absent or the check can skip.

Current code matches that trap. `tests/fwlive-upstream-cut.test.sh` always enforces cut shape and repo-relative `#:` paths. The Gap 5 fresh-scan half prints `ok` and exits 0 when `i18n-scan.pl` / `xgettext` are missing; CI has no luci checkout, so per-PR green does not mean the checked-in `.pot` was regenerated from source.

What is still ambiguous is the **outcome**:

- defer freshness with a revisit trigger of “upstream-cut / release, scanner present”; or
- fail closed on a documented sign-off path if the extractor is missing.

Without a filled ledger row (Implemented / Executed / Enforced / Manual plus one of the four outcomes), a later audit can still relitigate whether skip-as-pass is honest evidence or a silent gap. Suggest recording that disposition in the example, not only the trap.

**2. jshn / BusyBox compatibility**

“A bash-only stub is not equivalent evidence” is the right rule, but the example bundles two seams:

| Seam | What actually preserves it today | Escalation? |
| --- | --- | --- |
| libubox `jshn` JSON / `-eu` restore | Host, enforced: `tests/fwlive-jshn-compat.test.py` runs real matched pairs under `busybox sh` after `install-host-jshn.sh` | No, unless that gate starts skipping |
| BusyBox applets that differ from GNU (`stat -c`, `flock -w`, `nslookup`) | Host stubs can pin some (#232 `STAT=n`); applet-absent / installed-service failures still need QEMU (gap-2 `flock -w` residual) | Yes, when the host cannot omit the applet or observe the guest binary |

Reviewers can currently either send every ash/jshn question to QEMU, or claim host `busybox` covers applet-absent failures. Splitting the example would lock the intended answer: real jshn pairs stay a host gate; QEMU is for the applet / `ubus` / UCI boundary the host cannot preserve.

**3. “Layer 2” polling races**

This is the case most likely to produce split outcomes, mostly because of naming.

In `docs/developer/qemu-lab.md`, **Layer 2** is the unshipped #306 client backoff / visibility-pause work (“blocked on Layer 2”). The calibration text’s “epoch and bounded-work” reads instead as existing host contracts (`pollDataInFlight`, adaptive cap / Layer 1). Those already have host pins (`tests/fwlive-view-poll-guard.test.js`, `tests/fwlive-adaptive-cap.test.sh`).

A few distinctions the example should make explicit:

- Stub DOM in `tests/lib/load-fwlive-view.js` is **not** the mocked Playwright job. `tests/fwlive-view-poll-error.test.js` can assert banner `textContent` on a fake `document`; that does not prove LuCI `E()` / real event wiring.
- `npm run test:view` covers pause/resume CSS and chips; it does not cover hide/show with a poll outstanding, overlapping `pollData()`, or error→recovery in a real browser.
- Hide/show + outstanding requests is listed in the audit procedure, but the production Layer 2 client is not shipped. Is that “defer until the feature exists” or “current hole”? Either is defensible; the example should pick one.

Define “epoch” (in-flight guard vs adaptive bucket vs LuCI `poll.add`) and state stub DOM ≠ mocked Playwright ≠ QEMU LuCI lifecycle. Otherwise “browser or QEMU only for a distinct boundary” will be applied inconsistently.

**Worked-example request:** each of the three cases would be stronger as one filled ledger row (`risk | consequence | surface | assertion | artifact/fidelity | execution/enforcement | outcome | revisit`). If two people still diverge after that, the policy is not ready for the wider audit.

### Additional considerations

1. **Owner table.** `docs/developer/contributing.md` § Single source of truth does not list `test-approach.md`. This file now owns *what to test, which environment, how often*. `build-and-test.md` should keep commands; `validation-matrix.md` should keep cells. Without an owner row, the next docs PR will restate policy in those files (the #137 class of drift).

2. **Ledger location.** The policy refers to “the coverage ledger” as if it exists. There is no file or named issue. For a single maintainer, “issue comment using this template” is a fine choice — but say so, or the first audit will invent a committed file.

3. **Two evidence vocabularies.** `security-review.md` already uses `host` / `lab` / `manual` proof classes. The new Implemented / Executed / Enforced / Manually verified states are better (they catch skip-as-pass), but an audit will otherwise maintain two ledgers. One sentence mapping them would be enough (`host` ≈ implemented+executed+enforced on `fwlive-test.sh`, `lab` ≈ QEMU executed, `manual` ≈ manually verified).

4. **“Normal per-PR path” vs CI.** The document lists host suite, link check, shell portability, and mocked-view. Required CI also includes `z3-verify`, `zizmor`, `actionlint`, and the jshn installer inside the host job. Either name those or point at `.github/workflows/fwlive-test.yml` so a later audit does not treat Z3 / workflow static analysis as optional.

5. **QEMU “scheduled / manual or workflow-dispatch”.** There is no scheduled QEMU workflow. Tag / `workflow_dispatch` smoke lives in `publish-packages.yml`. “Scheduled” currently overclaims. Drop it, or mark it as not implemented.

6. **Fourteen-day windows.** Two consecutive 14-day defect windows plus a trend window is the #240 audit method. As a standing completion criterion for every coverage review it is heavy for a one-developer package. Scoping it to project-wide audits (and leaving ordinary defect pins on the six-question table) would match how the repo actually works.

7. **Contributor workflow vs CI gate.** `contributing.md` still says LuCI/rpcd changes get a local `qemu-install` + manual check. That is compatible with “no routine per-PR QEMU CI” if called out as **Manually verified**, not as a missing CI job. A pointer would prevent the audit from filing “QEMU is not in PR CI” as a gap.

I would not block on expanding the calibration examples in this PR, but I would treat filling those three ledger rows (and adding the owner-table line) as the cheapest way to lock the policy before the later coverage audit uses it.

docs/developer/test-approach.md
```diff
@@ -0,0 +1,213 @@
...
+does not automatically belong in every pull request, and a cheap test is not
+automatically sufficient for every boundary.
+
+| Level | Use it for | Current policy |
+| --- | --- | --- |
+| Host unit/CLI | Parser, filter, classifier, shell helper, UCI-shape, error, timeout, and state-machine contracts that can be reproduced with deterministic fixtures or faithful dependency stubs. | Required in the normal host suite when the behavior is part of the per-PR contract. |
+| Browser with mocked services | Real browser DOM/rendering, LuCI-style `E()` behavior, poll and event wiring, visibility transitions, and theme behavior. The harness must state which LuCI services are mocked and which production semantics it preserves. | Use for UI contracts that host objects cannot observe; the required mocked-view job runs per PR. |
+| QEMU/OpenWrt | Installed-package seams: real LuCI dispatch, rpcd/ubus/UCI behavior, BusyBox ash and utilities, firewall/log pipeline, reload/uninstall behavior, and representative end-to-end recovery. | Keep focused smokes for release or upstream sign-off, scheduled/manual or workflow-dispatch; do not make QEMU a routine per-PR gate by default. |
```
lucas-albers-lz4: **Agreed direction — reject scheduled QEMU** (later review may still counterpoint).

There is no cron lane. Host + mocked-view run on every PR. The focused QEMU smoke belongs in publish/release CI (`v*` tag; the same workflow can be dispatched). Local QEMU stays a reproduction tool, not a timer.

Suggested replacement for this policy cell:

> Run the focused smoke in the publish/release CI job (tag push; the same workflow can be dispatched). Do not add a scheduled QEMU job, and do not make QEMU a routine per-PR gate. Local QEMU remains available to reproduce an installed-system failure.

The execution-tiers paragraph (L128–132) should say the same: QEMU runs as part of CI on the publish/release path, not on a calendar.

docs/developer/test-approach.md
```diff
@@ -0,0 +1,213 @@
...
+   to assertions and identify any proof that is only manual or local.
+3. Inspect dependency seams: BusyBox, ash, jshn, UCI, jsonfilter, firewall
+   helpers, LuCI DOM, and rpcd. Preserve the dependency whose semantics caused
+   the risk; stub unrelated services deterministically.
+4. Trace stateful and resource-sensitive behavior through failure and recovery:
+   poll failure/recovery, hide/show with requests outstanding, pause/resume,
+   cancellation, adaptive degradation, and return to normal operation.
+5. Cross-check escaped defects using two consecutive, explicitly dated
+   fourteen-day windows plus a longer trend window. Deduplicate by root cause;
+   history informs priority but does not waive severe risks.
```
lucas-albers-lz4: **Agreed direction — replace standing 14-day windows** (later review may still counterpoint).

The two-week windows were a one-time #240 mining method, not a cadence this repo uses. Calendar windows punish quiet periods and invent a second ritual on top of CI.

Suggested replacement for step 5:

> Cross-check escaped defects since the last recorded baseline: the previous coverage-review commit, or the last release tag (`v*`), whichever the ledger names. Deduplicate by root cause. History informs priority but does not waive severe risks. If that interval is too small to be informative, a longer lookback is optional context, not a second required window.

Ordinary work still does not mine history: every confirmed defect already needs a regression pin. Align the definition-of-done bullet (L195, “recent escaped bugs”) with that baseline, not a calendar.

lucas-albers-lz4 (commented):
## Maintainer direction (still open to later counterpoint)

Two policy replacements from the second-pass review are the current intended approach. Later reviewers can still argue against them; until then, treat these as agreed.

1. **No scheduled QEMU.** Tests run in CI, not on a calendar. Inline on the QEMU policy row.
2. **No standing 14-day windows.** Coverage reviews use escaped defects since a recorded baseline (previous coverage-review commit or last `v*` tag). Inline on the history step.

Ordinary per-PR work is unchanged: host suite + mocked-view in CI; every confirmed defect still gets a regression pin. These notes only change how QEMU is described and how a coverage audit mines history.

lucas-albers-lz4:
Second-pass criteria review applied in 71068a9.

Accepted:

- calibration cases now have explicit ledger rows and intended dispositions;
- `test-approach.md` is listed as the owner for test-selection and coverage-review policy;
- the default ledger location is an issue/PR comment using the documented template;
- security-review `host`/`lab`/`manual` proof classes are mapped without replacing that document’s ownership;
- the per-PR CI inventory points to the workflow and names jshn/BusyBox, Z3, and workflow static analysis;
- the host document shim, mocked browser job, and installed QEMU lifecycle are explicitly distinct;
- the contributor QEMU/manual check is identified as manual evidence, not a missing CI gate.

Kept the two maintainer-directed boundaries:

- no scheduled QEMU claim; release/tag or workflow-dispatch only;
- no standing two-week windows; audits use escaped defects since a recorded baseline, with short windows optional when useful.

The policy still does not begin the project-wide audit. The next step after criteria review is to apply the filled ledger to the shipped surfaces and produce a small ranked backlog.

docs/developer/test-approach.md
```diff
@@ -0,0 +1,239 @@
...
+The approach is based on the two-week bug reviews and the coverage plan in
+[#240](https://github.com/lucas-albers-lz4/fwlive/issues/240). The issue remains
+the historical decision record; this document is the durable policy. The
+implementation waves tracked by
+[#271](https://github.com/lucas-albers-lz4/fwlive/issues/271) are historical
+results, not a list of current gaps.
+
+## Core rule
+
+Add or strengthen a test when it protects a specific, consequential behavior
```
lucas-albers-lz4: 💡 STE 5.1/5.4: this first Core-rule sentence is 23 words (procedural limit 20) and puts the condition after the command. Suggest: "When a behavior is consequential and can fail plausibly, add or strengthen a test that detects that failure beyond the existing checks."

docs/developer/test-approach.md
```diff
@@ -0,0 +1,239 @@
...
+Review test evidence using these separate states:
+
+| State | Meaning |
+| --- | --- |
+| Implemented | The check exists in the repository. |
+| Executed | The check ran for the reviewed revision and its assertions were reached. |
+| Enforced | CI or a documented release/sign-off procedure requires it, and missing prerequisites fail that check rather than silently skipping it. |
+| Manually verified | A person performed a defined check that is not currently automated, with the date and artifact recorded. |
+
+Only the states supported by evidence should be claimed. An optional local
```
lucas-albers-lz4: 💡 "Only the states supported by evidence should be claimed." — `should` + passive (modal ladder: requirement becomes *must*; here imperative is cleaner): "Claim only the states supported by evidence."

docs/developer/test-approach.md
```diff
@@ -0,0 +1,239 @@
...
+This prioritization is a queue, not a permanent exemption. Reconsider a quiet
+area when its contract, dependency, supported platform, threat model, or
+failure evidence changes.
+
+## Current execution tiers
+
+The normal per-PR path is the host suite, real-jshn compatibility and BusyBox
+shell checks, link checking, the required mocked-LuCI view job, Z3 verification,
+and workflow static analysis (`zizmor` and `actionlint`), as defined in
+[`fwlive-test.yml`](../../.github/workflows/fwlive-test.yml). It should remain
```
lucas-albers-lz4: 💡 "It should remain fast enough for routine development." → "Keep the path fast enough for routine development." Same fix applies to `should be added only if` (line 178) and `The review ledger should have at least:` (line 181 → `The review ledger has at least:`).

docs/developer/test-approach.md
```diff
@@ -0,0 +1,239 @@
...
+
+This project uses defect-driven, risk-based testing. The goal is credible
+evidence for important behavior, not a target percentage of lines or tests.
+This approach records the criteria for deciding what to test, where to test it,
+and how often to run it. It is intended for a single developer maintaining a
+package that must remain suitable for upstream OpenWrt review.
+
+The approach is based on the two-week bug reviews and the coverage plan in
+[#240](https://github.com/lucas-albers-lz4/fwlive/issues/240). The issue remains
+the historical decision record; this document is the durable policy. The
```
lucas-albers-lz4: 💡 STE Rule 8.1 bans the semicolon (write two sentences). This document uses it ~25 times, mostly in the Purpose, Core rule, and the tables. Example split for this line: "Issue 240 remains the historical decision record. This document is the durable policy." Same pattern fixes the bulleted outcomes list (lines 78–81) — end items with periods, drop the semicolons."

lucas-albers-lz4 (commented):
## Review — docs: establish defect-driven test approach

**Verdict: Approve with minor nits** (0 blocking, 4 style items — STE pass run per repo convention).

### Fact-check (all verified against the repo at this head)
- Per-PR tier list matches `fwlive-test.yml` jobs exactly (host suite incl. real-jshn install, BusyBox shell parity, linkcheck, mocked view job, z3-verify, zizmor, actionlint).
- QEMU-on-publish claim matches `publish-packages.yml` (tag push + `workflow_dispatch`, no scheduled QEMU anywhere).
- jshn compat row is accurate: `tests/fwlive-jshn-compat.test.py` runs matched real-jshn pairs under BusyBox `sh` and **asserts** on a missing install rather than skipping — consistent with the doc's own fail-closed `Enforced` definition.
- Poll guard / adaptive-cap host tests exist (`fwlive-view-poll-guard.test.js`, `fwlive-view-poll-error.test.js`, `fwlive-adaptive-cap.test.sh`).
- #306 is still OPEN and not in the shipped surface — the `Defer ... until the behavior is in the shipped surface` disposition is current.
- `host`/`lab`/`manual` cross-reference matches the security-review.md proof-class table.
- Linkcheck passes on this head: 410 internal links, 0 broken; 66 external, 0 failed. Closing-keyword trap clear (`Refs #240` only).

### Nits (inline)
STE violations: condition-after-command + 20-word limit in the Core rule, `should` x4, semicolons throughout (Rule 8.1). None block merge.

lucas-albers-lz4 (commented):
## Doc review — test-approach.md

**Verdict: no blocking issues.** Every checkable claim in the document was verified against the tree:

- Per-PR tier description matches `fwlive-test.yml` exactly: host suite, real-jshn install + compat pair, BusyBox `sh` parity, link check, required `test-view-mock` job, `z3-verify`, `zizmor`, `actionlint`.
- QEMU policy matches `publish-packages.yml`: `smoke-from-feed` runs on tag push (dispatch can skip), no schedule, not a per-PR gate.
- `scripts/qemu-install-fwlive.sh` exists and is referenced in `contributing.md` as the manual step the doc calls `Manually verified`.
- #306 is still open, so the calibration row's "defer until the behavior is in the shipped surface" disposition is current.
- STE pass verified as applied (commit be12c74): no sentence-separator semicolons remain, no residual `should` modals, core-rule sentence now 19 words.

### What bugs really matter — assessment

The priorities table derives correctly from #240's defect history: shell/BusyBox portability was 5 of 8 open bugs and stays first; view error-surfacing (#233/#234/#55 silent failure shapes) is second; the UCI zone-identifier cluster folds into row 1. The "What can fail?" consequence list covers the dominant observed modes — silent degradation, fail-open, misleading output. The doc also correctly identifies the one live violation of its own Enforced definition (msgid parity silently skips when `i18n-scan.pl` is absent) and routes it to fail-closed strengthening instead of claiming it is enforced. The document practices its own criteria.

### Minor findings

1. **POT calibration row wording** — "Catalog-to-POT consistency is checked": the actual #273 assertions are cut-tree `.pot` byte-identity and repo-relative `#:` refs, not a catalog-to-POT consistency check. The doc's own criterion ("name the assertion, artifact, environment, gate") applies to the calibration table too — name the real assertions.
2. **Missing calibration example: UCI zone model** — #240 ranks the zone-identifier mismatch (#239 `@zone[N]` vs `cfgXXXX`, #168 named vs anonymous WAN zone) as the recurring cluster after portability, and it is the canonical case of "preserve the real dependency semantics" (`uci changes` and `uci show` emit different identifiers). A calibration row for this seam would show the criteria applied to the repo's second-highest defect cluster. Optional; the priority table already covers the area.
3. **`xgettext` vs the scanner** — the tree's scanner is `i18n-scan.pl`; `xgettext` appears nowhere else in the repo's tooling. Name one or the other so a reader does not look for an `xgettext` dependency that does not exist.

lucas-albers-lz4:
Resolved the final “Doc review — test-approach.md” findings in commit 22eef24.

- Rewrote the POT/source-freshness calibration row to name the actual upstream-cut assertions: checked-in POT preservation, repository-relative `#:` references, and conditional `i18n-scan.pl` msgid parity.
- Removed the ambiguous `xgettext`/scanner wording and consistently identify `i18n-scan.pl` as the source scanner, including its unavailable/skip disposition.
- Added the optional UCI calibration row covering named and anonymous WAN zones, `@zone[N]` versus `cfgXXXX` identity, duplicate names, and unrelated staged changes, with the host-versus-installed boundary stated explicitly.

Validation: `./scripts/fwlive-linkcheck.sh` passed with 410 internal links and 0 broken.

lucas-albers-lz4:
@coderabbitai review

lucas-albers-lz4:
@coderabbitai rate limit

Issue #339, opened 2026-09-15
#306: Validate Layer 2 browser performance under 2,000-row load


lucas-albers-lz4:
Child issue of #306.

## Goal

Produce real browser evidence for the Layer 2 visibility/lifecycle work on the 2,000-row canned payload before deciding whether #306 needs the bounded server summary schema and client summary fallback.

## Acceptance criteria

- Run the existing LuCI view through Playwright/CDP with a 2,000-row payload.
- Exercise the degraded profile with CDP CPU throttling at 4x (and 6x if the harness supports it).
- Verify hidden-tab behavior pauses polling and visible-tab recovery resumes it with the expected RTT/backoff behavior.
- Measure and record:
  - per-frame render time, with the target remaining below 250 ms;
  - largest main-thread task;
  - heap growth during a 30-minute soak, with the target remaining below 50 MB.
- Capture the browser/version, device/profile, payload size, poll cadence, sample count, and raw/summary results in the lab notes.
- If the gate fails, identify the dominant cost and propose the smallest follow-up: server summary, client rendering change, or both.
- Link the resulting evidence back to #306; do not claim the acceptance item is complete from host-only renderer tests.

## Scope

This issue covers measurement and the smallest performance fix required by the measured result. It does not add the #307 weak-device card/dismissal UI.

lucas-albers-lz4:
Started the Layer 2 gate.

Short real-guest smoke (x86 OpenWrt 24.10.8 + Chromium 148, 4x CPU throttle, 10s soak, checked-in 2,000-entry fixture):

- raw payload: 2,000 entries
- rendered rows: 1,143
- hidden-tab polls during 2.5s: 0
- visible-tab resume: passed
- render commit → paint: max 66.6 ms
- largest main-thread task: 721 ms
- heap growth: 0 MiB over the short smoke

This is harness validation, not #306 sign-off: the guest Chromium build does not expose Emulation.setPageVisibilityState, so the visibility result used the report-labeled document-emulation fallback; the required 30-minute soak and true CDP visibility run remain pending. The 721 ms long task is the first performance signal to investigate before adding summary mode. Refs #306.

lucas-albers-lz4:
Heap instrumentation and 2-minute gate completed.

Run: real x86 OpenWrt 24.10.8 guest + Chromium 148, 4x CPU throttle, 2-minute soak, 2,000 raw fixture entries / 1,143 rendered rows.

[...]

PR #343, opened 2026-09-15
feat(#339): cap weak-device table rendering at 250 rows

lucas-albers-lz4:
## Summary

- Cap DOM rendering at 250 rows on devices identified by the server as weak, while the selected Limit continues to control buffering.
- Explain the cap in the status line and add German, Russian, and Simplified Chinese translations.
- Add Playwright/CDP performance-harness controls for row-limit and weak-device fixture runs, with request/report validation and focused RPC tests.
- Record the browser evidence and practical 250-row guidance in the user and QEMU documentation.

The harness does not change the selectable Limit options or buffer retention. The production view applies the display-only cap when `logging_status.weak_device === true`; failed status refreshes preserve the last known cap state.

Refs #339
Refs #306


lucas-albers-lz4:
## Single-model review — PR #343

2 files, +32/−4. Test harness + docs only, no shipped surface. CI: all 12 checks green (verified). Syntax check on the PR version passes. Verified against base `48b6bda`.

**Verdict: approve with one non-blocking suggestion.**

### Suggestion (test strength)
- **tests/fwlive-layer2-performance.mjs (readiness gate)** — the old gate required `>1000` rows at limit 2000; the new gate is `>= Math.min(limit, 100)`, i.e. only `>=100` for limits 250–2000. A stalled render (~100 rows with limit set to 2000) would now pass the gate and flow into the soak, making `visible_rows` incomparable across candidate runs. Consider `limit <= 100 ? limit : 1000` (or at least preserving `>1000` for the 2000 case) so the four comparison points in the docs are measured from a fully-rendered table.

### Verified good
- Default `2000` preserves prior behavior; validation (`isInteger` + allowlist) rejects garbage with a clear error.
- Allowlist `[25, 50, 100, 250, 500, 1000, 2000]` matches shipped `ROW_LIMIT_OPTIONS` in `constants.js` exactly; docs correctly scope the #339 comparison to the 250/500/1000/2000 subset.
- `display_row_limit` reporting: fixture mode records the selected limit; `REAL_POLL` mode reads the live UI value — correct.
- Docs explicitly state this is measurement plumbing, no shipped-behavior claim — appropriate caution for #339.
- Body uses `Refs #339 / Refs #306`, no closing keywords — no auto-close risk. Tabs preserved per repo convention.

Nothing security-relevant (no DOM sinks, no ACL/shell/rpcd surface).

*Reviewed by Hermes Agent (single-model). Zen MCR follows.*

lucas-albers-lz4:
# Zen MCR — lucas-albers-lz4/fwlive PR #343

> **Head SHA:** `a70a344` • **Reviewers:** muse-spark-1.3, mimo-v2.5, nemotron-3-ultra (all Zen free) • **Synthesized by:** orchestrator from 3/3 delivered reviews
> **Verdict:** COMMENT (approve with findings — test-harness only, no shipped surface; no Critical from any reviewer)

PR parameterizes `FWLIVE_PERF_ROW_LIMIT` in the Layer 2 browser perf harness (+18/−4) and documents the 250/500/1000/2000 comparison for #339 (+14 docs). CI all green.

## Findings

### Major

**1. Readiness gate weakened — `visible_rows` may be captured mid-render (A, B, C — 3/3, high confidence, orchestrator-verified)**
- **File:** `tests/fwlive-layer2-performance.mjs` (PR: wait-gate predicate; base was `length > 1000`)
- **Problem:** Base required `tbody tr` count `> 1000`; PR requires `>= Math.min(limit, 100)`, i.e. 100 rows for every documented candidate (250/500/1000/2000) and the 2000 default. `visibleRows` is snapshotted immediately after the gate, so a 2000-run can baseline at ~100 rows and proceed before the table fills.
- **Why it matters:** #339's whole point is comparing render cost *at each limit*; a 2000-run gated at 100 rows measures the same ~100-row render as a 250-run, corrupting `visible_rows`, `render_commit_to_paint_ms`, and largest-task comparisons. The gate also no longer detects a limit selection that failed to take effect (select reads 2000, table stuck at shipped default 100).
- **Suggested fix:** Gate on `>= limit` (fixture always has 2000 rows), or `>= (limit >= 1000 ? 1000 : limit)` to preserve the original full-population intent while permitting 25/50/100.

**2. `FWLIVE_PERF_ROW_LIMIT` validated then silently ignored in REAL_POLL mode (A, B — 2/3, high confidence, orchestrator-verified)**
- **File:** `tests/fwlive-layer2-performance.mjs` (parse/validate block + `if (REAL_POLL)` branch: 3s sleep, no `selectOption`)
- **Problem:** The env is parsed, allowlisted, and defaults to 2000, but the REAL_POLL branch never applies `selectOption` — the reported `display_row_limit` reflects the guest's live select value (default 100 or hash-persisted `limit=`), not the env. An invalid value still aborts a REAL_POLL run where it would never be used.
- **Why it matters:** `REAL_POLL=1 FWLIVE_PERF_ROW_LIMIT=500` reports `display_row_limit: 100` with no warning — a misleading result the docs never mention.
- **Suggested fix:** Apply `selectOption` in REAL_POLL too, or warn/reject when both are set, plus one doc sentence that the env only affects the fixture-intercepted run.

### Minor

**3. `display_row_limit` reports intent, not observation (A, B, C — 3/3, high/medium)**
- **File:** `tests/fwlive-layer2-performance.mjs` (fixture branch: `displayRowLimit = PERF_ROW_LIMIT` overwrites the line-above DOM read)
- **Problem:** The DOM `inputValue()` read becomes dead code in fixture mode; the report echoes the request. Safe today only because the gate asserts DOM value === limit.
- **Suggested fix:** Re-read `inputValue()` after `selectOption`, or move the read into the REAL_POLL branch.

**4. `inputValue()` read unguarded in REAL_POLL path (B, C — 2/3, medium/low)**
- **File:** `tests/fwlive-layer2-performance.mjs` (`Number(await ...inputValue())`)
- **Problem:** Empty/non-numeric yields `0`/`NaN` verbatim into `display_row_limit`. Unlikely for a `<select>`, but the fixture path guards and this one doesn't.
- **Suggested fix:** `Number.isFinite` guard, report `null` on failure.

**5. Allowlist duplicates shipped `constants.js` with no cross-reference (A, B — 2/3, high on fact)**
- **File:** `tests/fwlive-layer2-performance.mjs` (`PERF_ROW_LIMIT_OPTIONS`) vs `constants.js:13 ROW_LIMIT_OPTIONS` — identical lists today.
- **Suggested fix:** Comment pointing at `constants.js` ("keep in sync").

**6. Docs omit 25/50/100 and the 2000 default (A, B, C — 3/3, high on fact)**
- **File:** `docs/developer/qemu-lab.md` ("existing Limit options 250, 500, 1000, 2000")
- **Problem:** Reader may think only four values are legal, and the default + REAL_POLL no-op are unstated.
- **Suggested fix:** State the full allowlist, the 2000 default, and the REAL_POLL scope; phrase as "e.g. the candidate options 250/500/1000/2000".

**7. Comparison omits shipped default 100 — no current-behavior baseline (C only, medium)**
- **Problem:** Harness defaults to 2000 (correct — preserves prior hardcoded behavior) but the #339 candidate set excludes 100, the actual shipped default, so there is no baseline against current UX.
- **Suggested fix:** Include 100 in the documented candidates or state why it's excluded.

### Nits
- `Number()` accepts hex/whitespace (`'0x7D0'`, `' 500'`) past the allowlist check — harmless given `includes`, strictness optional (A, C).
- Doc example omits `FWLIVE_CPU_THROTTLE` though the text requires same-throttle comparison (A, low).

## Agreement matrix

| Finding | muse | mimo | nemotron |
|---|---|---|---|
| Gate weakening (Major 1) | ✅ Major | ✅ Major | ✅ Major |
| REAL_POLL no-op (Major 2) | ✅ Minor | ✅ Major | — (touched via NaN nit) |
| Intent-not-observation (Minor 3) | ✅ | ✅ | ✅ |
| Unguarded inputValue (Minor 4) | — | ✅ | ✅ nit |
| Allowlist duplication (Minor 5) | ✅ | ✅ | — |
| Doc omissions (Minor 6) | ✅ nit | ✅ | ✅ |
| Missing 100 baseline (Minor 7) | — | — | ✅ |
| No Critical | ✅ | ✅ | ✅ |

No reviewer disagreements on substance; only severity calibration (REAL_POLL no-op: Major vs Minor).

## Things reviewers could not verify (all noted runtime behavior)
Whether the table renders 2000 rows synchronously vs incrementally under `FWLIVE_CPU_THROTTLE=4` — determines if the 100-row gate undercounts in practice. Needs a live QEMU guest run; worth one short soak at limit 2000 to confirm before the #339 comparison.

## Verified good (orchestrator ground-truth + reviewer consensus)
- Allowlist `[25, 50, 100, 250, 500, 1000, 2000]` matches shipped `ROW_LIMIT_OPTIONS` exactly; validation (`isInteger` + `includes`, raw input echoed in error) correctly rejects `NaN`/`'500.5'`/out-of-set.
- Default 2000 preserves prior hardcoded `'2000'` behavior; `waitForFunction` arg passing (`PERF_ROW_LIMIT` as Playwright arg) is correct.
- "Measurement plumbing only" scope claim is accurate — no `constants.js` change; body uses `Refs`, no auto-close risk.

*Zen MCR: 3/3 reviewers delivered, findings ground-truthed against base `48b6bda` / head `a70a344`. Full reviews in run transcript.*

lucas-albers-lz4:
Applied the later review feedback in 9053b28.

- The fixture readiness gate now passes the selected limit and expected visible-row count together: 1000 for the 1000/2000 fixture candidates, or 250 when weak-device mode is enabled.
- An explicit FWLIVE_PERF_ROW_LIMIT is now honored in REAL_POLL mode, and display_row_limit is re-read and validated from the rendered select rather than echoing intent.
- The harness allowlist and QEMU-lab documentation now cover the full shipped range (25, 50, 100, 250, 500, 1000, 2000), the 100-row shipped baseline, fixture defaults, and real-poll override behavior.

Validation: full ./scripts/fwlive-test.sh passed; normal 2000-row live run observed 1143 rows; weak-device 2000-row run observed 250 rows with a 237 ms largest task and no long tasks >=250 ms; explicit real-poll limit 500 was observed in the report. The existing uci/get access-denied pageerror remains a lab harness condition and does not prevent completion. Refs #339

lucas-albers-lz4:
Applied the remaining review finding in 14a3996.

The fixture route now intercepts both fwlive.poll and fwlive.logging_status, including standalone and batched requests. This makes FWLIVE_PERF_WEAK_DEVICE=1 reliable instead of allowing the initial logging_status call to reach the guest. Request-ID matching and successful ubus-result validation remain in place for the real-poll observation path.

Added focused routing/matching coverage for standalone logging_status, rules + logging_status batches, reordered replies, and non-target rules requests. Full ./scripts/fwlive-test.sh passes. No shipped security/XSS findings were identified by the prior Astra pass. Refs #339

lucas-albers-lz4:
## Review (head `14a3996`)

CI is green. CodeRabbit's two findings (preserve weak-device cap after status errors; observe guest `weak_device` in real-poll reports) are already addressed.

This is no longer harness-only. Head ships a **250-row DOM cap** when `logging_status.weak_device === true`, while Limit still controls the in-memory buffer. The view split is the right one: `applyFetchedEntries` still uses `rowLimit`; `filteredRows()` slices with `displayRowCap()`. Status text is `textContent` only. `loadLoggingStatus()` only updates `weakDevice` on a successful reply, and poll start waits on that status call.

### Blockers

None for the shipped cap or the status-error preservation.

### Should fix before merge

**PR title and body are stale.** Title is still `test(#339): parameterize display-limit performance runs`, and the body still says this is measurement plumbing that does not change shipped Limit behavior. Head adds the production cap, i18n, user docs, and a security-review row. Retitle to something like `feat(#339): cap weak-device table rendering at 250 rows` and describe the shipped behavior plus the harness/evidence.

**`docs/developer/qemu-lab.md` contradicts the product change.** It still says the work does not change shipped Limit options and that the weak-device toggle is harness-only. Production now reads `logging_status`. The 250/500 measurement notes are useful; the “plumbing only” paragraph is not.

### Non-blocking

**2000-limit readiness still gates at 1000 rows.** For a non-weak run, `expectedVisibleRows` is `1000` whenever `PERF_ROW_LIMIT >= 1000`. A 2000 vs 1000 Limit comparison can still soak a ~1000-row table. That does not undermine the 250-vs-500 evidence that chose the cap (those sizes fill completely). It does mean the documented 2000 candidate is not guaranteed to be a 2000-row paint. If that comparison still matters, wait for `>= PERF_ROW_LIMIT` in fixture mode (fixture IDs are unique and length is 2000).

**Tests only cover `weak_device: true` plus a later status failure.** Add a `weak_device: false` / missing-flag case so a strong device at Limit 2000 still returns 1000 rows from `filteredRows()`, and that a string `"true"` does not trip the strict boolean.

**Harness magic `250` is not tied to `WEAK_DEVICE_DISPLAY_ROW_CAP`.** `ROW_LIMIT_OPTIONS` already has a keep-in-sync comment; the display cap should too.

**Dead local in the fixture route:** `loggingStatusIds` is computed and unused; `isFwliveFixtureRequest()` already covers that.

**UX nit:** live `compactCountText()` uses capped `filteredRows().length` as “matching”, so a weak device at Limit 2000 can read `250 matching · 2000/2000 stored` even when more rows match filters. The suffix explains the cap; paused mode already counts the full buffer. Follow-up, not a hold.

### Already in good shape

- Strict `weak_device === true` matches the fail-closed server boolean.
- Status-error preservation is tested.
- RPC matching for batched/reordered `logging_status` replies is covered.
- Real-poll reports observed guest `weak_device`, or `null` if unseen.
- i18n + user-doc sentence for the cap are present.
- No ACL or HTML-sink change.

Approve after the title/body and qemu-lab wording match the shipped 250-row cap. The 1000-row gate and missing non-weak test can land here or in a fast follow-up.

lucas-albers-lz4:
Applied the later Grok review feedback in bb2924e.

- Retitled the PR and rewrote the body to describe the shipped 250-row DOM cap, buffering behavior, harness coverage, and evidence.
- Corrected qemu-lab.md so it distinguishes production weak-device behavior from synthetic fixture control.
- Added strict weak_device=false, missing, and string-value tests; the cap remains fail-closed.
- Tied the harness cap literal to constants.WEAK_DEVICE_DISPLAY_ROW_CAP with an explicit sync comment and removed the unused route local.

The 2,000-selection readiness threshold remains 1,000 rows for the non-weak 1,000/2,000 cases because the 2,000-entry fixture yields about 1,143 classified/displayable rows; requiring 2,000 rendered rows would make that candidate impossible to measure. The compact-count wording is a separate UX refinement: the cap is already explicitly reported in the suffix, and changing the count semantics would add a new display-contract/i18n decision beyond this evidence slice. Refs #339

lucas-albers-lz4 (commented):
## Hermes review — delta `bb2924e` (vs `14a3996`)

Verdict: approve. All four items from the prior round are addressed; no new shipped surface in this delta.

### Verified
- Title/body now describe the shipped 250-row DOM cap (`feat(#339)`). Body uses `Refs`, no auto-close risk.
- `qemu-lab.md` corrected: production cap vs synthetic fixture toggle now distinguished; "measurement plumbing only" paragraph gone.
- Fail-closed coverage added: `weak_device: false`, missing flag, and string `"true"` all assert `weakDevice === false` with full 2000 cap. Re-ran `fwlive-view-layer2-backoff.test.js` and `fwlive-layer2-performance-rpc.test.mjs` locally — pass.
- Harness cap literal tied to shipped constant via `WEAK_DEVICE_DISPLAY_ROW_CAP` + sync comment; dead `loggingStatusIds` local removed from the fixture route (remaining use at the response observer is the live one). `node --check` clean on both harness files; tabs clean.

### Notes (non-blocking)
- `test` CI job still pending at review time; everything else green. Don't merge until it lands.
- 2000-selection readiness gating at 1000 rows stands with the stated reason (2000-entry fixture yields ~1143 classifiable rows, so `>= 2000` would be unmeasurable). Acceptable, with the consequence that 2000 vs 1000 comparisons read off `visible_rows`, which the report already records.
- Compact-count wording left as-is (cap reported in suffix). Agreed this is a separate UX/i18n decision, out of scope here.

*Reviewed delta `14a3996..bb2924e` against head `bb2924e`; local unit tests re-run, full QEMU soak not repeated.*

lucas-albers-lz4:
## Hermes review — follow-up on delta `bb2924e`

Two non-blocking notes from the prior round still stand — omitted from my delta review by mistake, not fixed by the delta:

- **`docs/developer/qemu-lab.md` still bakes in one lab run's numbers** (50.9 ms p95, 175.7 ms max, 217 ms largest task, Chromium 148/KVM 4x throttle). Environment-specific and will rot. Suggest scoping as a dated example run or moving the table to #339, keeping the doc to method + fields.
- **`fwliveRpcReplyForRequest` still matches IDs with strict `===`** (`tests/lib/fwlive-perf-rpc.mjs`, untouched by this delta). Fine today since IDs round-trip the same type, but fragile across string/number ID producers. Consider `String(id)` comparison.

Neither blocks; verdict stays approve (pending `test` CI).

lucas-albers-lz4:
Applied the two remaining non-blocking follow-ups from the latest review in `066dc49`.

- `docs/developer/qemu-lab.md` now labels the 2026-09-15 measurements as dated, environment-specific example evidence and directs future runs to repeat the documented method on current hardware/browser combinations.
- `fwliveRpcReplyForRequest()` now matches JSON-RPC IDs using normalized string values, with focused coverage for a numeric request ID and string response ID.

Validation: focused tests, `npm run lint:js`, `git diff --check`, and the full `./scripts/fwlive-test.sh` all pass. The PR remains open and unmerged; CI is rerunning for `066dc49`. Refs #339

lucas-albers-lz4:
@coderabbitai review

Issue #340, opened 2026-09-15
Design: make JavaScript easier to reason about through clear ownership and explicit state


lucas-albers-lz4:
## Objective

Make fwlive's JavaScript easier to understand, change, and review by improving responsibility boundaries, state ownership, names, and local clarity.

> Beautiful code: a reader can understand a behavior, predict its effects, and change it safely by reading one coherent piece of code with a small amount of surrounding context.

**Design/planning issue.** Deliverable: agreed boundaries, examples, and staged slices — not every refactor shipped. Implementation children open only after the gates below.

### Review status (folded)

| Pass | Verdict | Notes |
|---|---|---|
| Multi-reviewer design review (2-of-3) | Approve + conditions | Conditions folded below; dropped-lens questions answered |
| Disposition + design review | Approve (updated body) | Cross-issue bounds confirmed |
| Independent design review | Approve with changes | API/prefs-test wording gaps folded below |

Baseline: current master `24c18ed` after the #306 Layer 2, #339 evidence, #344 harness, and small-device documentation slices. Historical code links below remain review anchors; WIP is not baseline truth. Default row limit is **100** (not 96).

**Preserve:** [`buffer` helpers](https://github.com/lucas-albers-lz4/fwlive/blob/60788d91965c8e47fdf35b057fe9dffacbd6627d/openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/fwlive/buffer.js#L52), [`normalizeEntry()`](https://github.com/lucas-albers-lz4/fwlive/blob/60788d91965c8e47fdf35b057fe9dffacbd6627d/core/fwlive-log.js#L293), host/state/callbacks renderers, text-node sinks, plain LuCI JS.

## Ownership map

| Issue | Owns | Not |
|---|---|---|
| #306 | Request **behavior** (in-flight, epoch/visibility, cadence, adaptive health, evidence) | Module extraction, naming, prefs sequencing shape |
| #338 | Auto/Manual **contract** (keys, hash precedence, wire, banners, init-before-fetch) | View refactor that implements sequencing |
| **#340 (this)** | **Shape** (coordinator interface placement, prefs-resolve step, names, local clarity, classifier representation) | Competing behavior/contract text |
| #339 | Browser **evidence** for row/render cost | Owning readability refactors |
| #336 | Whether shipped comments keep `#N` tracker ids | Broader comment quality |

[...]

Issue #338, opened 2026-09-15
Design follow-up to #306: Auto fetch budgeting with numeric row limits and bounded Manual override


lucas-albers-lz4:
### Summary

Refs #306.

Define an Auto/Manual fetch-budget setting alongside the numeric display limit. Auto should use a predictable request budget derived from the display limit, with the existing server-side adaptive controller deciding how much work to permit.

This is a **design/specification issue for independent review**, not an implementation request. Its deliverable is an agreed UX, request/reply contract, compatibility policy, and implementation test plan.

### Existing behavior

- Layer 1 already adjusts the server's line cap using measured request processing duration. The hard ceiling is 2000 raw lines; hot behavior can reduce the cap to 250 without exceeding a smaller request.
- Layer 2 already adjusts polling **cadence** using client RTT and surfaces adaptive/shedding/truncation status. It does not independently control line counts.
- The hot floor and the cool cap are the same 250-line value, reached through different controller paths.
- The numeric `Limit` currently controls retained/displayed rows and indirectly controls fetch sizing:
  - Live: `min(max(rowLimit × 4, 100), 2000)`. This derivation is static: the client never adjusts the multiplier from response metadata.
  - Paused fetches: 2000.
- The request parser maps empty, zero, and non-digit input to the default 50 and caps digit input at 2000. The adaptive compute path floors the applied limit at 1. Only valid digit strings reach 1.
- `rowLimit` must remain numeric: buffering, slicing, URL state, and browser persistence depend on that assumption.
- A separate session-dedup memory cap (`Math.max(rowLimit × 2, 2000)`) also derives from the display limit. Fetch-budget modes do not change it.

The opportunity is to explain existing automatic behavior and provide an explicit request-budget override without introducing another adaptive line-count controller.

### Proposed controls and semantics

| Control | Meaning |
|---|---|
| **Limit** | Existing numeric 25–2000 display/retention setting. Unchanged by this design. |
| **Fetch budget: Auto** | Default. Derives the requested raw-line budget from Limit. The server applies its adaptive cap. |
| **Fetch budget: Manual** | Exposes **Maximum raw lines per fetch**, a separate numeric setting chosen from `{25, 50, 100, 250, 500, 1000, 2000}` — the same discrete option set as Limit, reusing its validation style. Server protection still applies. |

[...]

Issue #307, opened 2026-09-07
Discussion: weak-device card vs first-run consent card — structure and dismissal model


lucas-albers-lz4:
## Purpose

Discussion proposal, not an execution plan. Nothing here is a directive. The weak-device install card from #306 overlaps the shipped first-run consent card. The two interact on card structure, dismissal persistence, and new config surface.

**Why this issue:** Maintainer wants security and functionality trade-offs reviewed together before locking the model, so the decision moved out of #306. That issue keeps the settled half: server-side `weak_device` detection on `logging_status`.

---

## Grounded facts (verified at `23ebfc8897`, tip of #303 branch)

1. **Consent card ships.** Trigger (`shouldShowLoggingConsent()`): `logging_status` loaded AND `!wan_log` AND no blockers AND not permanently dismissed AND not session-dismissed. Buttons: Enable WAN logging (also persists dismissal), Not now, Don't show this again checkbox, firewall zones link.
2. **Dismissal = localStorage + session flag.** `persistConsentDismissed()` writes to localStorage; `consentDismissedSession` is in-memory view-instance only (a reload resets it — it is *view-instance lifetime*, not a guaranteed login- or browser-session lifetime).
3. **Enabling logging suppresses the card cross-browser — with a caveat (corrected).** `wan_log` is checked first, so after `enable_wan_logging` the card is suppressed from any browser *while logging stays on*. But `wan_log` is **current operational state, not a permanent acknowledgment**: the enable action *also* writes a localStorage dismissal, and disabling logging does not clear it — so after enable→disable the card does **not** re-arm in that browser. If re-arming matters, either `localStorage.removeItem` on disable or document a manual re-arm path.
4. **`logging_status` reply fields:** wan_zone, wan_log, wan_log_limit, nf_log_ipv4, nf_log_ipv6, ready, blockers, warnings.
5. **ACL scope:** read = rules/poll/resolve/logging_status, write = enable/disable_wan_logging. Write methods own UCI-write + firewall-reload path with rollback.
6. **No `/etc/config/fwlive`** exists yet (#306 adds one with `option adaptive`).
7. **#306 adds `weak_device`** bool to `logging_status`, computed in rpcd from `/proc/meminfo` + `/proc/cpuinfo` (browser cannot read `/proc`; `ubus system board` carries no memory/core count). No ACL change.

[...]

Issue #209, opened 2026-08-29
Planning: upstream luci-app-fwlive to openwrt/luci


lucas-albers-lz4:
## Goal

One planning issue for the first PR of **`luci-app-fwlive`** into [`openwrt/luci`](https://github.com/openwrt/luci) at `applications/luci-app-fwlive/`. This is the culmination of the package-readiness work (issue #16 and follow-ons). **Do not open the luci PR until the prep checkboxes on the doc are done and a CodeRabbit round on the linked draft PR has landed.**

Canonical checklist (same content, easier to diff): [`docs/developer/upstream-openwrt.md`](docs/developer/upstream-openwrt.md).

## Walk-risks stay here (not separate issues)

LuCI review risks are **proposed PR-body answers**, not independent workstreams. Confirm or amend **in this issue**. Do not pre-fix code unless CodeRabbit or a luci reviewer shows a nack.

| Topic | Position |
|-------|----------|
| WAN-log write path | Session ACL does **not** grant `uci` or `log.read`; rpcd toggles as root; opt-in; `prerm` restores `/etc/fwlive/wan-log-baseline`. |
| Custom rpcd vs `ubus log.read` | Intentional. Sessions must not get `log.read`. |
| Package size / `__selftest` | Leave CLI-only selftests in the shipped scripts; explain if asked. |
| `prerm` | Undoes WAN `log=1`; fail-open so uninstall cannot strand. |
| Translations | `.pot` only in the first luci PR; Weblate after merge. |
| `PKG_VERSION` 0.1.36 | Keep in the cut; honest snapshot. |
| Out-of-tree heuristic | **New** in-tree app (not the `luci-app-https-dns-proxy` redirect list). fwlive stays development home; binary feed stays. |
| Generated files | Snapshots from this repo. Do not vendor Node generators into luci. |

The rpcd `list` phantom `"backend":"unknown"` is **prep**, not a PR-body excuse — fix in this repo before the cut.

## Prep (this repo)

- [ ] Drop `"backend":"unknown"` from rpcd `list` JSON
- [ ] Extend `scripts/upstream-cut.sh` (`rm -rf` locale dirs, file-count, README / GENERATED / `constants.js` rewrites; optional drop `fwlive.css`)
- [ ] `./scripts/fwlive-test.sh` + `./scripts/validate-baseline.sh`
- [ ] After luci-shaped copy: `./build/i18n-scan.pl` POT (must include menu title + ACL description); copy back; same luci commit

[...]

PR #348, opened 2026-09-16
feat: add Auto/Manual fetch budgets (#347)

### LINKED ISSUE #347: Implement frozen #338 Auto/Manual fetch budget contract
## Objective

Implement the frozen Auto/Manual fetch-budget contract from #338, using #340's current-tree coordinator and state-ownership boundaries.

This is the first implementation child after the #338 design freeze. Keep the existing `Limit` display/retention semantics and Layer 1/Layer 2 behavior intact.

## Scope

- Resolve RPC-affecting preferences (`Limit`, poll mode, Manual maximum) before the first fetch and poll registration, while leaving filter-widget hash application in `addFooter`.
- Add the bounded Manual fetch-budget control and persistence/hash rules from #338; Auto remains derived from configured Limit.
- Add the server `effective_limit` success-only field with a success/error-path audit, including `filter_tempfile_failed` and `classifier_missing`.
- Add the two budget status templates, filling state, and legacy/malformed metadata handling with text-node rendering and i18n updates.
- Preserve one active request plus one coalesced follow-up, visibility epoch behavior, resume merge, and Layer 2 cadence/backoff.
- Add the frozen host/browser contract tests, including combined Limit+visibility and Manual/filling vectors.
- Update `docs/openwrt-fwlive-schema.md` and `docs/developer/security-review.md` in the same change.

## Ownership boundaries

- #338 owns persistence keys, URL precedence, wire fields, tristate adaptive interpretation, `effective_limit`, and banner templates.
- #340 owns coordinator shape and state ownership; do not perform an unrelated broad refactor in this child.
- #306 owns server adaptive behavior and existing Layer 2 semantics; do not add a second adaptive controller.
- #307 remains design-only; no weak-device card, write ACL, or UCI surface.
- #209 is out of scope.

## Required gates

- Preferences-before-first-RPC is tested against production `load` → `render` → `addFooter` order.
- Auto boundary, Manual clamp/seeding, URL order, explicit Auto write-through, paused mode changes, visibility/filling, zero-growth, errors, legacy replies, and malformed `effective_limit` vectors are covered.
- Every untrusted value reaches the DOM as text nodes. No ACL changes.
- Run the full `./scripts/fwlive-test.sh` and documentation linkcheck before handoff.

Refs #338
Refs #340
Refs #306
Refs #307

lucas-albers-lz4:
## Summary

Implements the frozen fetch-budget contract from #338 using the coordinator boundaries from #340.

- Adds Auto and bounded Manual raw-line budgets with persistence and URL precedence.
- Resolves RPC-affecting preferences before the first poll.
- Adds success-only effective_limit metadata and error-path handling.
- Adds budget status reporting, malformed/legacy metadata handling, and paused-buffer filling state.
- Preserves serialized refresh, visibility epochs, resume merge, and Layer 2 cadence behavior.
- Updates schema, user, security, and i18n documentation.
- Adds coverage for paused changes, hidden Limit catch-up, filling, pagehide disposal, and filter failure paths.

## Validation

- ./scripts/fwlive-test.sh
- npm run test:links
- 416 internal links and 69 external URLs checked with no failures or warnings

## Scope

No ACL, UCI, weak-device write surface, or #209 work is included.

Refs #338
Refs #340
Refs #306
Refs #307

Closes #347

## Summary by CodeRabbit

* **New Features**
  * Added Auto and Manual fetch-budget modes for live firewall logs.
  * Added configurable maximum raw-line limits in Manual mode.
  * Added status details for requested, server-applied, and returned message counts.
  * Added “buffer filling” status while paused during Manual fetching.
  * Added preference persistence through the URL and browser storage.
  * Successful requests now report effective server limits.

* **Documentation**
  * Documented fetch budgets, adaptive metadata, polling behavior, and view limits.

* **Translations**
  * Added translation entries for the new controls and status messages.


lucas-albers-lz4:
Astra review found two confirmed P2 issues to address before merge:

1. Budget changes currently trigger an immediate or queued extra poll. The fetch-mode and Manual-maximum handlers call requestPoll(), which can add a fetch immediately after an active request and bypass the configured cadence/backoff. This conflicts with the frozen #338 behavior that changes take effect on the next scheduled poll. Proposed fix: update preferences and controls without requesting a poll, then add idle and in-flight tests under slow cadence.

2. The first switch to Manual can use an outdated Auto-derived value. Startup always initializes manualFetchLines, so selecting Manual after changing Limit to 25 can choose 250 instead of the current Auto budget of 100. Proposed fix: track whether a Manual value was explicitly restored or selected, and seed from the current Auto budget on first entry to Manual. Add a regression for changing Limit before that first selection.

Astra found no additional confirmed issues in the rpcd response changes, XSS boundaries, i18n, or security behavior. The jshn-dependent check was skipped because jshn is unavailable.

The findings are also recorded on issue #347.

openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js
```diff
@@ -1312,6 +1535,28 @@ return view.extend({
...
+		const mode = ev && ev.target ? ev.target.value : '';
+		if (constants.FETCH_MODE_OPTIONS.indexOf(mode) < 0) return;
+		this.fetchMode = mode;
+		if (mode === 'manual' && !this.manualFetchLines)
+			this.manualFetchLines = this.snapManualFetchLines(this.autoFetchLines());
+		this.saveFetchMode();
+		if (mode === 'manual') this.saveManualFetchLines();
+		this.updateStreamControlsUi();
+		this.updateHash(this.readFilters());
+		if (!this.paused) this.requestPoll().catch(function () {});
```
lucas-albers-lz4: Astra P2-1: this `requestPoll()` (and the same call in `onManualFetchLinesChange`) inserts an extra fetch on live budget changes.

Idle: it starts immediately and bypasses Layer 2 cadence. In-flight: `requestPoll()` queues the coalesced follow-up, which re-issues after the current RPC. #338 says a mid-flight setting change does not re-issue; the in-flight reply is accepted and the new budget applies on the next scheduled poll. Help text also says Manual changes the budget, not the interval.

Drop both `requestPoll()` calls. Persist, update chrome/hash, and let the existing timer pick it up. Add idle + in-flight tests under slow cadence. Keep Limit’s fetch; that is the handler #338 says not to copy.

openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js
```diff
@@ -1312,6 +1535,28 @@ return view.extend({
 		else this.renderRows(true);
 	},
 
+	onFetchModeChange(ev) {
+		const mode = ev && ev.target ? ev.target.value : '';
+		if (constants.FETCH_MODE_OPTIONS.indexOf(mode) < 0) return;
+		this.fetchMode = mode;
+		if (mode === 'manual' && !this.manualFetchLines)
```
lucas-albers-lz4: Astra P2-2: this reseed never runs after load. `resolveRpcPreferences()` already snaps `manualFetchLines` from the Auto budget whenever storage is empty, including while the user is still in Auto. After that the value is always truthy.

First visit Limit=100 seeds 250 in memory; changing Limit to 25 then switching to Manual persists 250 instead of current Auto 100.

Track whether a Manual value was restored or explicitly chosen. On first entry to Manual with no such value, snap from `autoFetchLines()` at that moment. Add a regression for Limit change before the first Manual selection.

openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js
```diff
@@ -194,27 +202,161 @@ return view.extend({
...
 				const n = parseInt(val, 10);
-				if (isFinite(n) && constants.ROW_LIMIT_OPTIONS.indexOf(n) >= 0) {
-					this.applyRowLimit(n);
-					this.saveRowLimit();
-				}
+				if (isFinite(n) && constants.ROW_LIMIT_OPTIONS.indexOf(n) >= 0) hashLimit = n;
+				continue;
+			}
+			if (key === 'poll') {
+				hashMode = constants.FETCH_MODE_OPTIONS.indexOf(val) >= 0 ? val : 'auto';
```
lucas-albers-lz4: Invalid `poll=` is coerced to `'auto'` and then write-throughs below, wiping stored Manual mode. Invalid `limit=` / `maxraw=` are ignored instead.

#338: hash values are exact lowercase `auto`/`manual`; a **valid** URL value wins and is persisted; anything else falls back without clobbering storage.

Only set `hashMode` when `val` is in `FETCH_MODE_OPTIONS`. Leave stored mode alone for `poll=foo`, `poll=MANUAL`, or `poll=`.

lucas-albers-lz4 (commented):
## Review

Agree with the [Astra P2s](https://github.com/lucas-albers-lz4/fwlive/pull/348#issuecomment-5702683916). Both are real #338 contract misses. One additional persistence bug, then this is in good shape.

The prefs-before-first-RPC split, paused Auto/Manual budgets, success-only `effective_limit`, tristate adaptive, pagehide disposal, and text-node banners all look right. CI is green.

GitHub will not accept `REQUEST_CHANGES` on the author's own PR, so this is posted as a comment review. Treat the three inline notes as merge-bar items.

### Astra P2-1 — extra poll on budget change — **confirmed**

`onFetchModeChange` / `onManualFetchLinesChange` call `requestPoll()` whenever the table is not paused.

- Idle + slow cadence: the change fires immediately and bypasses Layer 2 backoff.
- In-flight: it queues the coalesced follow-up, which re-issues after the current RPC instead of accepting that reply and waiting for the next scheduled poll.

That conflicts with #338: a setting change must not re-issue, and Manual changes the fetch budget, not the polling interval. The paused test only asserts `requestPoll` is skipped while paused; it does not cover idle or in-flight under a 5s cadence.

Fix: persist + update chrome/hash only. Let the existing poll timer pick up the new budget. Add the idle and in-flight slow-cadence tests Astra named. Limit can keep its own fetch; #338 calls that out as the pattern **not** to copy.

### Astra P2-2 — stale Manual seed — **confirmed** (independent of Astra)

`resolveRpcPreferences()` snaps `manualFetchLines` on every first visit, including while still in Auto. `onFetchModeChange` only reseeds if that value is falsy, which it never is after load.

First visit, Limit 100 → in-memory Manual 250; change Limit to 25; switch to Manual → persists 250 instead of current Auto 100. Seed (or re-snap) on first Manual entry when nothing has been restored or chosen yet, and add the Limit-then-Manual regression.

### Additional — invalid `poll=` write-through

Invalid `limit=` / `maxraw=` are ignored. Invalid `poll=` (`poll=foo`, `poll=MANUAL`, `poll=`) is coerced to Auto **and saved**, which wipes a stored Manual preference. #338: only a valid URL value write-throughs; anything else falls back without clobbering storage. Treat a present-but-invalid `poll` like invalid `maxraw`.

### Non-blocking

Astra’s all-clear on rpcd/XSS/security holds. Remaining gaps are coverage, not new product bugs:

- `fwlive-rpcd-security.test.js` never asserts `effective_limit` is omitted on the six error replies (`jsonfilter_missing` is missing from that matrix). The merge helper test covers the flag, not the live rpcd fixtures. The security-review row overclaims that coverage.
- Frozen vectors still untested: `effective_limit` `0` / `2001` / `250.5`; hot-cooldown full reply with truncation and no applied number; stored Manual=25 + URL `limit=100` only.
- `de`/`ru`/`zh_Hans` copy English for the new strings. The i18n gate only checks non-empty msgstr.
- After the `!grew` return, the last `fillingBuffer` assignment is dead.

**Merge bar:** P2-1, P2-2, and the invalid-`poll` persist bug.

lucas-albers-lz4:
Astra and Grok findings have been reviewed and addressed in commit d5fb888.

Confirmed fixes:
- Removed immediate requestPoll calls from fetch-mode and Manual-maximum changes. Budget changes now persist and update the UI/hash, while the existing timer applies them on the next scheduled poll without bypassing Layer 2 cadence or queuing an extra request.
- Added explicit tracking for restored or selected Manual values. First Manual entry now seeds from the current Auto budget, including after a Limit change.
- Invalid poll hash values are ignored without overwriting stored mode.

Coverage added:
- Idle and in-flight budget changes under slow cadence.
- Limit change before first Manual selection.
- Invalid poll persistence.
- effective_limit values 0, 2001, 250.5, and malformed strings.
- effective_limit omission across the read/filter error matrix.

Validation:
- ./scripts/fwlive-test.sh passed.
- npm run test:links passed: 416 internal links and 69 external URLs, no failures or warnings.

The remaining review notes were non-blocking observations: translated catalog entries retain the project’s existing English-placeholder convention, and no additional product change was justified for the dead assignment observation.

lucas-albers-lz4:
@coderabbitai rate limit

lucas-albers-lz4:
Delta review of `d5fb888` (P2-1, P2-2, and invalid-`poll` fixes).

Verified, no issue:

- P2-1 fix holds: both budget handlers persist and update chrome/hash with no fetch trigger. The Limit handler keeps its own fetch, as intended.
- P2-2 fix holds: the explicit Manual tracking closes the stale-seed path. Traced every `manualFetchLines` writer at the new head — load seed, hash restore, first Manual entry, manual change — no other path clobbers it, and the first-entry snap reads the already-updated Limit.
- Invalid `poll=` is ignored without touching storage, and duplicate `poll` keys resolve order-independently (only a valid value assigns).
- The `jsonfilter_missing` fixture reproduces the exact body `fwlive-log-filter.sh` emits, and the `effective_limit` omission assert now covers the error matrix.

New finding (minor, test strength):

- `testBudgetChangesRespectCadence` asserts no fetch (`idleCalls` 0 / `inFlightCalls` 1) but never asserts post-state: no check on mode, stored value, or hash after the change. Its paused sibling (`testPausedBudgetChangesDoNotFetch`) does assert those. As written, the cadence test would still pass if the handlers stopped persisting. Suggest asserting `fetchMode` / `manualFetchLines` plus storage/hash after the change, mirroring the paused test.

lucas-albers-lz4:
Addressed the additional review note in commit 76a6ead.

The slow-cadence budget-change test now asserts both sides of the contract:
- no immediate or queued extra RPC;
- fetch mode and Manual maximum update correctly;
- the updated mode/value are persisted and reflected in the hash.

Focused fetch-budget tests, JavaScript lint, formatting, and diff checks pass. The full CI gate is running on the new head.

lucas-albers-lz4:
@coderabbitai review

PR #349, opened 2026-09-16
test: add fetch-budget qualification harness

lucas-albers-lz4:
## Summary

- Extend the reusable Layer 2 browser gate with explicit Auto/Manual fetch-budget controls.
- Seed display/fetch preferences through the URL before the first poll for deterministic qualification runs.
- Capture every `fwlive.poll` raw-line request and report expected versus observed budgets under `fetch_budget`.
- Add focused RPC request-capture coverage and runbook examples.

## Validation

- `node --check tests/fwlive-layer2-performance.mjs`
- `node tests/fwlive-layer2-performance-rpc.test.mjs`
- `npm run lint:js`
- `npm run lint:format`
- `git diff --check`

The full repository gate reaches an unrelated pre-existing checkout-mode failure: the AWK file is checked out as 0664 while its test expects 0644; this branch adds no mode change. The live guest was unavailable for the browser run.

Refs #339
Refs #306

## Summary by CodeRabbit

- **New Features**
  - Added configurable automatic and manual fetch modes for performance runs.
  - Added fetch-budget validation and reporting for polling requests.
  - Added support for including row limits and fetch settings in performance URLs.
  - Real-visibility runs now require an explicit fetch mode.

- **Tests**
  - Added coverage for fetch-budget calculations, request parsing, URL generation, and budget matching.
  - Added validation for malformed or missing polling data.


lucas-albers-lz4:
## Review

Harness-only change for #339 fetch-budget qualification: URL-seed Auto/Manual, capture `addresses[0]`, report `fetch_budget`. RPC parser tests pass. The documented matrix (`FWLIVE_PERF_ROW_LIMIT=500` plus explicit `auto`/`manual`) can work, but the new report can still describe a mixed-budget run as a clean expected value.

### 1. Fixture default Limit is still applied after the first poll

`fwlivePerformanceUrl()` only puts `limit=` in the hash when `FWLIVE_PERF_ROW_LIMIT` is set. Fixture mode still always `selectOption`s `PERF_ROW_LIMIT`, which defaults to **2000**.

So the existing default gate:

```sh
FWLIVE_CPU_THROTTLE=4 FWLIVE_SOAK_MS=1800000 ./scripts/qemu-layer2-performance.sh
```

still starts at the shipped Limit 100 (Auto raw **400**), then jumps to 2000 (Auto raw **2000**). That was already true; this PR now reports it under `fetch_budget`:

- `expected_raw_lines` is computed from the **post-select** Limit → `2000`
- `requested_raw_lines.unique` will be `[400, 2000]`
- the process still exits 0

`onRowLimitChange` always `requestPoll()`. Leaving the post-load `selectOption` in place after URL seeding can also fire an extra poll even on the documented `ROW_LIMIT=500` path, depending on whether Playwright emits `change` for an already-selected value.

**Fix:** in fixture mode, always hash the effective Limit (including the 2000 default). Skip `selectOption` when `#fwlive-limit` already matches. Keep `selectOption` only as a fallback when the guest lacks hash support.

### 2. Expected vs observed is never checked

`fetch_budget` records `expected_raw_lines` and `requested_raw_lines` but does not compare them. `FWLIVE_ENFORCE=1` still ignores this field.

For a qualification harness, a mismatch is the failure mode: stale `fwlive-poll-mode=manual` on a headed CDP profile, hash dropped on redirect, or the warmup mix in (1). At least fail (or warn loudly) when `requested_raw_lines.unique` is not exactly `[expected_raw_lines]` after the URL-seeded prefs are applied.

When `FWLIVE_PERF_FETCH_MODE` is unset, `expected_raw_lines` still uses the Auto formula (`min(max(limit*4, 100), 2000)`). That is wrong if the guest stored Manual. Either require Auto/Manual for qualification, or set `expected_raw_lines` to `null` unless the mode was explicit.

### 3. Coverage gap vs the new contract

`fwlivePollRequestedLines` is tested. `fwlivePerformanceUrl()` and the Auto expected formula are not. Those are the bits that have to stay aligned with `resolveRpcPreferences()` / `autoFetchLines()` in `fwlive.js` (`limit`, `poll`, `maxraw`, `* 4` / floor 100 / cap 2000). A small host assertion that the default fixture URL includes `limit=2000`, and that Auto expected matches `autoFetchLines()`, would have caught (1).

---

Nits: duplicate `PERF_MANUAL_LINES_OPTIONS` vs `PERF_ROW_LIMIT_OPTIONS`; hard-coded `2000` instead of a keep-in-sync comment next to `FETCH_LINES_MAX`. Neither is blocking.

I would not treat a `fetch_budget` JSON blob as #339 evidence until (1) and (2) are fixed or the report clearly flags mixed `unique` values.

lucas-albers-lz4:
# Multi-Model Review — lucas-albers-lz4/fwlive PR #349

> **Head SHA:** `13ebd1b2c1937b315b910e8b09bd50a433334b29`  •  **Reviewers:** glm-5.3-flash, deepseek-v4.1-flash, qwen3.7-max  •  **Synthesized by:** opencode/deepseek-v4-pro

## TL;DR
This PR extends the Layer 2 performance harness with URL-seeded Auto/Manual fetch-budget controls, per-poll capture of `fwlive.poll` `addresses[0]` values, and an `expected_raw_lines` vs `requested_raw_lines` comparison in the JSON report, plus docs and an RPC unit test. The implementation is clean, the env-var validation is thorough, and the RPC helper mirrors the app's single-element `addresses` contract correctly. The consensus problem is the `fetch_budget.expected_raw_lines` field: all three reviewers independently concluded it does not model what the polls actually request, in two distinct ways. The remaining items are duplicated-constant drift risk, preference-persistence leakage, and RPC test coverage gaps. Recommend fixing the `expected_raw_lines` modeling before this evidence is used for the #339 qualification.

### 🔴 Critical
None.

### 🟡 Major

- **Finding:** Default invocation computes `expected_raw_lines` from the final display limit, contradicting the first poll's request
- **File:** `tests/fwlive-layer2-performance.mjs:595-599` (via `fwlivePerformanceUrl()` at `:283-293`)
- **Reviewers who flagged:** B (high confidence); C's suggested fix (`expected_raw_lines: null` when mode is null) touches the same line
- **Problem:** `fwlivePerformanceUrl()` only seeds `limit` into the URL hash when `FWLIVE_PERF_ROW_LIMIT` is explicitly set. In the flagship documented run (row limit unset), the first poll uses the stored/default limit (100 → `addresses[0]="400"`), the harness then selects 2000, and `expected_raw_lines` is computed from the final `displayRowLimit` (2000). The report therefore emits `expected_raw_lines: 2000` against `requested_raw_lines.unique: [400, 2000]` — internally contradictory qualification evidence.
- **Suggested fix:** Always push the effective `limit` into the URL, or record the row limit live at the first request and derive `expected_raw_lines` from that value.
- **Confidence:** high

- **Finding:** Auto-mode `expected_raw_lines` ignores the paused/filling-phase polls that request `FETCH_LINES_MAX` (2000)
- **File:** `tests/fwlive-layer2-performance.mjs:595-599` (app side: `fwlive.js:321-326`, `autoFetchLines()` at `fwlive.js:255`)
- **Reviewers who flagged:** A (medium), C (medium)
- **Problem:** `requestedFetchLines()` returns `constants.FETCH_LINES_MAX` (2000) while the view is paused/buffer-filling, not the `displayRowLimit * 4` auto budget. The first 1–2 polls will request 2000 regardless of `displayRowLimit`, so `requested_raw_lines.unique` will contain a value that never equals `expected_raw_lines`. A qualification analyst comparing the two may flag filling-phase polls as budget violations when they are by design.
- **Suggested fix:** Report `expected_raw_lines` per phase (running vs paused), or set it to `null` when `PERF_FETCH_MODE` is null and document that auto runs include filling-phase polls at `FETCH_LINES_MAX`; alternatively split the summary into filling vs steady-state.
- **Confidence:** medium

### 🟢 Minor

- **Finding:** Duplicated shipped constants without "Keep in sync" comments (auto formula `100`/`4`/`2000`; `PERF_MANUAL_LINES_OPTIONS` duplicate of `PERF_ROW_LIMIT_OPTIONS`)
- **File:** `tests/fwlive-layer2-performance.mjs:119` (options list), `:599` (formula)
- **Reviewers who flagged:** A (high), B (high)
- **Problem:** This file already relies on explicit "Keep in sync with `constants.*`" comments for exactly this pattern (lines ~121-122). The new formula and option list are unlabeled copies of shipped constants and will silently drift when `constants.js` changes, corrupting reports.
- **Suggested fix:** Reuse `PERF_ROW_LIMIT_OPTIONS` for manual lines and add a sync comment referencing `constants.FETCH_LINES_MAX` / `autoFetchLines()`.

- **Finding:** Stale browser preference can persist across real-visibility runs and mislabel the workload
- **File:** `tests/fwlive-layer2-performance.mjs:595-599`; docs `docs/developer/qemu-lab.md:466-470`
- **Reviewers who flagged:** B (medium); A raised the same point as a nit (docs-only)
- **Problem:** The app persists `fwlive-poll-mode`/`fwlive-manual-lines` (`fwlive.js:240-252`, `saveFetchMode()` at `:305-308`). In `FWLIVE_PERF_REAL_VISIBILITY` mode the external context is reused, so a prior manual run's stored mode survives a later run that omits `FWLIVE_PERF_FETCH_MODE`, while `expected_raw_lines` is still computed as auto. The docs claim "a stale browser preference cannot change the workload," which only holds for manual.
- **Suggested fix:** Require `FWLIVE_PERF_FETCH_MODE` in real-visibility mode, or record/clear the resolved mode for unqualified runs; add one docs line noting auto also persists.

- **Finding:** RPC request-capture edge cases untested
- **File:** `tests/fwlive-layer2-performance-rpc.test.mjs:37-62`
- **Reviewers who flagged:** A (nit), B (high), C (high)
- **Problem:** Only batched payloads with non-empty `addresses` and the empty-args case are covered. Missing: `addresses: []`, single-object payload with addresses, malformed JSON, and non-string `addresses[0]`. The browser path also never asserts `observed == expected`.
- **Suggested fix:** Add the missing asserts to the RPC unit test (e.g., `fwlivePollRequestedLines(JSON.stringify({... addresses: []}))` → `[]`).

- **Finding:** Manual-lines validation error message is confusing when fetch mode is unset
- **File:** `tests/fwlive-layer2-performance.mjs:118`
- **Reviewers who flagged:** C (high)
- **Problem:** Setting only `FWLIVE_PERF_MANUAL_LINES=500` triggers the guard `PERF_FETCH_MODE !== 'manual' && PERF_MANUAL_LINES !== null`, but the message "requires manual fetch mode" doesn't make clear that omitting the mode entirely also trips it.
- **Suggested fix:** Change message to `'FWLIVE_PERF_MANUAL_LINES requires FWLIVE_PERF_FETCH_MODE=manual'`.

### 💡 Nits

- `tests/fwlive-layer2-performance.mjs:286` — `encodeURIComponent(PERF_ROW_LIMIT)`/`(PERF_MANUAL_LINES)` coerce numbers via implicit `toString()`, inconsistent with the view's `%s` formatting (C, no change needed).
- `tests/fwlive-layer2-performance.mjs:396` — `counts.requestedLines` is camelCase while all report fields are snake_case (B).
- `tests/fwlive-layer2-performance.mjs:98-103` — `FWLIVE_PERF_FETCH_MODE` is not trimmed/lowercased, so `auto ` throws; consistent with existing env parsing, low value (B).

## Reviewer agreement matrix

|---|---|---|---|---|
| `expected_raw_lines` default-run mismatch (final vs first-poll limit) | — | ✓ | ✓ | B leads; C's fix (null when mode null) overlaps |
| `expected_raw_lines` ignores paused/filling-phase 2000 polls | ✓ | — | ✓ | A & C flag same line, B silent |
| Duplicated constants / missing sync comments | ✓ | ✓ | — | Both high confidence |
| Stale browser preference persistence | ✓(nit) | ✓(minor) | — | A docs-only, B wants mode required |
| RPC test coverage gaps | ✓(nit) | ✓ | ✓ | All three, varying severities |
| Manual-lines error message clarity | — | — | ✓ | C only |

- **Severity of the persisted-preference issue:** Reviewer A treated it as a nit (one docs line), while B escalated it to a minor (require the mode in real-visibility mode, since the reused external context can silently mislabel a run). B's framing is the stronger of the two; whether a docs-only fix suffices is a human call given this is a dev-only harness.
- **Root cause of the `expected_raw_lines` mismatch:** A and C identified the paused/filling-phase `FETCH_LINES_MAX` behavior; B identified the first-poll limit seeding in the default run. These are complementary, not contradictory — both are true and both live on the same ternary. The fix should address both, not choose one.

- **`mode: null` conflation with auto in the report ternary** (`tests/fwlive-layer2-performance.mjs:595-599`): when `FWLIVE_PERF_FETCH_MODE` is unset, `fetch_budget.mode` reports `null`, but `expected_raw_lines` still falls through to the auto formula (non-null). The report can therefore never show `mode: null` alongside `expected_raw_lines: null` — structurally inconsistent, and the clearest articulation of what C's fix suggestion is driving at. (C hinted at this in the fix; A and B did not state it directly.)
- **`requested_raw_lines.count` vs `counts.polls` divergence** (`:402-408`): `counts.requestedLines` is pushed on every poll request regardless of interception, while `counts.polls` only increments in `REAL_POLL` mode. In default interception runs `requested_raw_lines.count` will be large while `polls` stays 0; likely intended, but a reader comparing the two fields may be confused. Worth a comment or a doc note.

- [ ] Fix `expected_raw_lines` to model actual request state: record the row limit live at the first poll and derive expected from it, and account for (or annotate) paused/filling-phase `FETCH_LINES_MAX` polls (Major #1 + #2).
- [ ] Set `expected_raw_lines` to `null` when `PERF_FETCH_MODE` is null so the report is internally consistent (synthesizer note).
- [ ] Add "Keep in sync with `constants.FETCH_LINES_MAX` / `autoFetchLines()`" comments and reuse `PERF_ROW_LIMIT_OPTIONS` for manual lines (Minor).
- [ ] Add the missing RPC edge-case asserts (`addresses: []`, single-object payload, malformed JSON, non-string `addresses[0]`) (Minor).
- [ ] Clarify the manual-lines validation message (Minor).
- [ ] Address or document the persisted-preference leakage in real-visibility runs (Minor).

- Reviewer A (glm-5.3-flash): Thorough and well-grounded — verified the app-side contract (`fwlive.js:893` single `addresses` entry, `autoFetchLines()`, `saveFetchMode()`), validated env-var matrix, and ran the RPC unit test. Hedged on lint (deps not installed) and the live browser run. Sound but the paused-phase finding is medium-confidence.
- Reviewer B (deepseek-v4.1-flash): Most rigorous on the report-correctness path — caught the default-run first-poll seeding issue that A and C missed, with high confidence, plus a sharper persistence concern. Also hedged on lint and the live guest; raised a good "couldn't verify" on whether the fixture ignoring `addresses[0]` is intended.
- Reviewer C (qwen3.7-max): Careful and precise — reproduced the paused-phase issue with exact `fwlive.js` line refs, added a concrete fix for `expected_raw_lines: null`, and flagged the confusing error message. Hedged on whether `scripts/qemu-layer2-performance.sh` forwards the new env vars (env vars inherit by default, so likely a non-issue).

## Process notes
- Wall time: ~360s (estimate from file mtimes; not precisely instrumented)
- Reviewer failures: none (all three `.err` files empty; all three reports produced)
- Diff was truncated: no (263 lines, full)
- Total cost: not tracked in this pass

lucas-albers-lz4:
Addressed the review findings in commit 1930792.

- Always URL-seed the effective display Limit, including the default 2000, and skip the post-load selector change when it already matches; this removes the initial 100/400 warm-up and avoids an unnecessary extra poll.
- `expected_raw_lines` is now null when fetch mode is unset, and explicit Auto/Manual runs compute the expected value through a shared tested helper.
- Explicit runs now report `matches_expected` and fail if captured poll requests are mixed, malformed, or different from the expected budget.
- Reused `PERF_ROW_LIMIT_OPTIONS` for Manual values and added sync comments for the shipped option sets and Auto formula.
- Real-visibility runs now require an explicit fetch mode so persisted browser preferences cannot silently change the workload.
- Added request-capture edge coverage for empty addresses, single-object payloads, non-string values, and malformed JSON.
- Documented that this gate covers the running phase; it does not click Pause, while hidden-tab polling is handled separately.

Focused RPC tests, syntax checks, shipped-surface lint/format, and diff checks pass. The PR remains draft until a live guest is available for the actual browser qualification matrix.

lucas-albers-lz4:
Resolved both CodeRabbit findings in commit e192066.

- Batched JSON-RPC payloads now increment `polls` by the number of poll calls, in both fixture and real-poll paths. Poll calls are counted even when `addresses` is missing, so malformed requests cannot evade the budget check.
- Budget reporting now retains invalid raw values in `requested_raw_lines.invalid_requested_raw_lines`; valid integer summaries remain unchanged, and the matcher still fails the configured run on invalid/mixed values.
- Added focused tests for batched poll counting, malformed/negative raw values, and the existing budget matcher behavior.

Focused syntax, RPC, lint/format, and diff checks pass. The new CI run is in progress.

lucas-albers-lz4:
Astra final general review of PR #349 at e192066b612231f6b7a8a4e980d9c34c00451464 found no confirmed actionable findings. It reviewed the diff against master, the batched poll accounting, malformed and negative address reporting, budget validation, tests, and documentation. No files were changed. The only limitation noted is that live QEMU/browser qualification was not available, so end-to-end timing and visibility behavior remain unverified.

Issue #350, opened 2026-09-17
Implement #340 coordinator disposal and first-poll preference ordering


lucas-albers-lz4:
Child issue of #340.

## Goal

Implement the first current-tree slice of the #340 JavaScript ownership contract: make request lifecycle disposal explicit and resolve RPC-affecting preferences before the first poll.

## Scope

- Preserve the landed requestPoll coalescing, one-active-request rule, epoch stale-reply discard, visible catch-up, and resume-merge obligation.
- Add an explicit coordinator/disposal seam for pagehide: invalidate active work, remove cadence and visibility registrations, settle or ignore active work safely, and prevent queued work or buffer/DOM application after disposal.
- Add the combined Limit-plus-visibility in-flight lifecycle test.
- Resolve the #338 RPC-affecting preferences before poll registration and the initial request, while keeping filter widget hash writes in addFooter.
- Add the first-poll assertion against fwlive.poll addresses[0].
- Keep buffer application, render scheduling, hostname resolution, and DOM outside the coordinator boundary.
- Keep classifier representation, comment-policy work, and opportunistic renderer refactors out of this slice.

## Acceptance criteria

- The two lifecycle vectors recorded on #340 are covered by behavior-level tests.
- Existing Layer 2 lifecycle and fetch-budget tests remain green.
- First request uses the resolved stored/hash Limit and later #338 fetch-budget context.
- No shipped DOM or ACL security invariants regress.
- Required JS/lint/format/full test checks pass.

Refs #340
Refs #338

lucas-albers-lz4:
Initial implementation slice pushed on `feat/350-coordinator-dispose` at `512a185`.

- `disposeCoordinator()` now immediately settles active and queued waiters, invalidates the epoch, and ignores the later RPC completion.
- Added the combined Limit + visibility in-flight lifecycle test.
- Full `./scripts/fwlive-test.sh`, JS lint, formatting, syntax, and diff checks pass.

Refs #340

lucas-albers-lz4:
Completed by merged PR #351 (`74fc7dd`), including explicit pagehide disposal, active/queued waiter settlement, stale-reply protection, and the combined Limit plus visibility lifecycle coverage. All CI and review gates passed.

PR #351, opened 2026-09-17
test(#350): cover coordinator disposal lifecycle

lucas-albers-lz4:
## Summary

- Make coordinator disposal settle active and queued waiters immediately.
- Invalidate the epoch and ignore late RPC completion after pagehide.
- Add coverage for Limit plus visibility changes during an in-flight request.
- Verify the visible catch-up uses the current Auto fetch budget.

## Validation

- `node tests/fwlive-view-fetch-budget.test.js`
- `./scripts/fwlive-test.sh`
- `npm run lint:js`
- `npm run lint:format`
- JavaScript syntax and diff checks
- Luna pre-PR review: clean
- Bugbot was unavailable in this environment
- Live QEMU/browser validation was unavailable

Refs #350
Refs #340

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved status view cleanup when leaving or hiding the page.
  * Prevented stale polling results from appearing after the view is disposed or becomes hidden.
  * Ensured queued refreshes resume correctly with the latest row limit when the view becomes visible again.
  * Resolved pending refresh requests reliably during page navigation.


tests/fwlive-view-fetch-budget.test.js
```diff
@@ -411,11 +461,15 @@ async function testPagehideDisposesCoordinator() {
...
+	const queued = v.requestPoll();
+	assert.strictEqual(calls, 1, 'second request must be queued before pagehide');
 	h.dispatchPagehide();
-	release();
+	await queued;
 	await loading;
+	assert.strictEqual(v.pollDataInFlight, false, 'pagehide must settle active waiters');
+	release();
+	await sleep(10);
 	assert.strictEqual(v.entries.length, 0, 'pagehide must discard active reply');
```
lucas-albers-lz4: This assertion cannot catch a stale apply. The mock still returns `{ log: [] }`, so `entries.length === 0` holds whether or not `fetchEntries` discards the reply.

The new `await queued` / `await loading` before `release()` already proves waiter settlement. To prove discard, return a real row from `fwlive.poll` (same pattern as `testLimitAndVisibilityDuringInFlightRequest`) and assert that row never lands in `v.entries` after `release()`.

lucas-albers-lz4 (commented):
Production disposal looks right: `pagehide` now settles active and queued waiters immediately, bumps the epoch, and `finishPollRequest` cannot start follow-up work on the detached run.

One test gap below. The Limit+visibility in-flight vector is solid. Not blocking on the commit subject, but `test(#350)` undersells the shipped waiter-settlement change.

tests/fwlive-view-fetch-budget.test.js
lucas-albers-lz4: Confirmed. The fixture now returns a real firewall row (id 913), so the post-release empty-buffer assertion proves the stale reply was discarded rather than merely observing an empty payload. Added in 06c09b2; focused lifecycle tests and lint pass.

lucas-albers-lz4:
@coderabbitai review

lucas-albers-lz4:
# Multi-Model Review — lucas-albers-lz4/fwlive PR #351

> **Head SHA:** `512a1859301b8848d0fdf5957672e829817b006d`  •  **Reviewers:** qwen3.7-max, minimax-m3, kimi-k2.7-code  •  **Synthesized by:** deepseek-v4-pro
> **Verdict:** APPROVE

## TL;DR
This PR makes `disposeCoordinator()` eagerly settle both active and queued poll-request waiters when the view departs (pagehide), relying on the existing epoch/`pollRequestPromise` guards to discard any late RPC completion. It adds a lifecycle test for Limit + visibility changes during an in-flight request and strengthens the pagehide test to cover both the active and queued disposal paths. All three reviewers agree the change is correct, narrowly scoped, and free of critical/major issues; the only feedback is comment wording and optional test-assertion hardening. Safe to merge.

### 🔴 Critical
None.

### 🟡 Major
None.

### 🟢 Minor
- **Finding:** Code comment misattributes which guard discards the detached run.
- **File:** `openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js:1175-1177`
- **Reviewers who flagged:** B
- **Problem:** The new comment says "the bumped epoch makes its result stale and `finishPollRequest` ignores the detached run." In fact `finishPollRequest` ignores the run via the `pollRequestPromise !== run` identity check; the epoch check lives in `fetchEntries`/`runPollRequest`. Reads as if `finishPollRequest` itself inspects the epoch.
- **Suggested fix:** Reword to attribute the two guards correctly, e.g. "...the bumped epoch makes its result stale; `finishPollRequest` ignores it via the `pollRequestPromise !== run` check."
- **Confidence:** medium
- **Synthesizer note:** I agree with B's framing but am re-ranking this from "minor" to borderline nit — it is a pure comment-accuracy/clarity issue with zero behavioral impact. Kept under Minor since B explicitly rated it so and the wording could mislead future maintainers tracing the defense-in-depth story.

### 💡 Nits
- **Finding:** Test relies on `calls === 1` to prove hidden transition didn't dispose the coordinator, without asserting the coordinator is still alive.
- **File:** `tests/fwlive-view-fetch-budget.test.js:387`
- **Reviewers who flagged:** A
- **Problem:** If `disposeCoordinator` were triggered during `setHidden(true)`, `calls` would still be 1, so the assertion would pass for the wrong reason. End-to-end the later `calls === 2` catch-up assertion still fails in that scenario, so the test is sound, but intent is not self-documenting.
- **Suggested fix:** Add `assert.strictEqual(v.coordinatorDisposed, false, ...)` after line 387.
- **Confidence:** low

- **Finding:** Test doesn't assert the queued waiter actually exists when pagehide is dispatched.
- **File:** `tests/fwlive-view-fetch-budget.test.js:385-388`
- **Reviewers who flagged:** B
- **Problem:** After the Limit change + hide, the test asserts `calls === 1` and `entries.length === 0` but never asserts `v.pollRequestQueued === true`, so a future refactor that silently drops queued waiters in `stopPollingForHidden`/`bumpPollEpoch` could still pass as long as no second RPC fires.
- **Suggested fix:** Add `assert.strictEqual(v.pollRequestQueued, true, ...)` before `release()`.
- **Confidence:** low

- **Finding:** Active-waiter drain loop duplicates the queued-waiter drain loop a few lines below.
- **File:** `openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js:1178-1182`
- **Reviewers who flagged:** B
- **Problem:** Verbatim-ish duplication; could become a maintenance hazard if one path gains error handling.
- **Suggested fix:** Leave as-is (a shared helper would obscure the linear teardown narrative).
- **Confidence:** low

## Reviewer agreement matrix

|---|---|---|---|---|
| Comment misattributes which guard discards the run (epoch vs identity) | — | ✓ | — | Only B; rated Minor |
| Test should assert coordinator still alive (line 387) | ✓ | — | — | Only A; low confidence |
| Test should assert queued waiter exists before pagehide | — | ✓ | — | Only B; low confidence |
| Drain-loop duplication | — | ✓ | — | Only B; suggests leaving as-is |

- **No substantive disagreements.** All three independently concluded the production change is correct and free of critical/major issues. The only divergence is which optional test assertion to add: A suggests asserting `coordinatorDisposed === false`; B suggests asserting `pollRequestQueued === true`. Both are complementary and non-conflicting (they harden different aspects of the same test region); the author could take both. B rates the comment-wording issue "Minor" while A and C treat the code as clean — this is a severity call on a non-functional nit, not a disagreement about correctness.

- **Active waiters are `resolve()`d, not `reject()`d.** `disposeCoordinator` resolves the drained active waiters (and the existing code resolves queued ones), so a caller of `requestPoll()`/`load()` observes a successful (settled) promise even though no data was applied. This is consistent with the pre-existing queued-waiter drain and is benign given the epoch/identity guards and the fact that the view is disposed; callers should not interpret settlement as "data received." Worth a one-line note in the comment only if the team wants to lock in the semantics. (Low confidence; flagging only because no reviewer noted the resolve-vs-reject choice.)

- [ ] (Optional) Reword the `disposeCoordinator` comment to attribute the identity check vs the epoch check correctly (B, Minor).
- [ ] (Optional) Add `assert.strictEqual(v.coordinatorDisposed, false, ...)` to `testLimitAndVisibilityDuringInFlightRequest` (A, nit).
- [ ] (Optional) Add `assert.strictEqual(v.pollRequestQueued, true, ...)` before `release()` in the same test (B, nit).
- [ ] (Optional) Confirm whether resolving (vs rejecting) drained waiters is the intended contract and, if so, note it in a comment.

- Reviewer A (qwen3.7-max): Thorough. Ran `node tests/fwlive-view-fetch-budget.test.js` directly (12 subtests pass); flagged that `scripts/fwlive-test.sh` exits early at the jshn test and that eslint/prettier were not installed. Cited specific line numbers and walked the epoch-then-settle ordering. High confidence in the no-critical/major verdict.
- Reviewer B (minimax-m3): Thorough. Cited concrete line numbers (`fwlive.js:900`, `:2093`, test `:348`, `:464`), verified both defense-in-depth guards, and noted lint/prettier clean. Hedged on browser/reload ordering and production promise-holding behavior.
- Reviewer C (kimi-k2.7-code): Solid but thinner. Confirmed logic sound, tests pass, lint/format clean; cited the same guard lines. No findings. Noted jshn missing and no live QEMU/browser validation.

## Process notes
- Wall time: ≈13 minutes for the three review passes (prompt files 00:45 → review `.md` files 00:58), plus synthesis.
- Reviewer failures: none (all three produced reports).
- Diff was truncated: no (2 files, +65/−2, full patch present in working dir).
- Total cost: not precisely tracked; synthesis session ≈$0.01; review sessions not summed but comparable order of magnitude.

lucas-albers-lz4:
@coderabbitai review

lucas-albers-lz4:
Reviewed the follow-up multi-model findings and applied the confirmed/hardening items in 61a8847:

- clarified that the epoch makes the late fetch stale and the cleared promise makes finishPollRequest reject the detached run by identity;
- asserted the coordinator remains live on visibility hide;
- asserted the Limit refresh remains queued before the hidden request is released.

Focused lifecycle tests, JS lint, formatting, and diff checks pass. Triggering the next CodeRabbit pass on this updated head.

lucas-albers-lz4:
@coderabbitai review

lucas-albers-lz4:
CodeRabbit review completed on 61a8847 with no actionable comments. Its only tooling limitation was Biome rejecting the LuCI module’s top-level return shape; repository node syntax, ESLint, Prettier, focused lifecycle tests, and full ./scripts/fwlive-test.sh validation pass. No further code changes are needed from this review.

Issue #352, opened 2026-09-17
Implement #340: extract the poll coordinator boundary


lucas-albers-lz4:
Child issue of #340.

## Goal

Extract request coordination from the LuCI view into a small, dependency-injected coordinator module so request lifecycle ownership is explicit and testable without DOM or buffer knowledge.

## Scope

- Move ownership of one active request, one coalesced follow-up, request epoch, visibility transitions, cadence registration, and disposal into a coordinator module.
- Preserve the behavior landed through #306 and #351:
  - no overlapping requests;
  - hidden requests are discarded;
  - one visible catch-up request;
  - Limit/Pause/Resume refreshes coalesce;
  - pagehide settles active and queued waiters and prevents late application;
  - cadence changes re-register polling without creating an RPC.
- Keep the view responsible for fetch/normalize/buffer application, rendering, table pause state, resume merge, hostname resolution, and DOM.
- Use injected callbacks/adapters for the request runner, visibility state, poll registration, and epoch-aware result handling.
- Add module-level behavior tests for the coordinator state machine and keep the existing view lifecycle tests.
- Do not change #338 wire/persistence semantics, classifier representation, or renderer behavior.

## Acceptance criteria

- Every poll request start is owned by the coordinator and discoverable in one module.
- Existing lifecycle and fetch-budget tests remain green.
- Coordinator tests cover active request plus queued follow-up, hidden/show catch-up, combined Limit plus visibility, pagehide disposal, cadence changes, and no post-disposal request.
- No buffer, DOM, hostname, or ACL logic enters the coordinator.
- Full repository validation passes.

Refs #340
Refs #306
Refs #351

lucas-albers-lz4:
Completed in PR #353, which was squash-merged to master as `aedb5b87462fae8bb3134db368505bca3642dfcb`. The coordinator now owns request lifecycle, visibility/epoch handling, cadence registration, coalescing, and disposal while buffer, rendering, hostname, and DOM responsibilities remain in the view.

Refs #353
Refs #340

PR #353, opened 2026-09-17
refactor(#352): extract poll coordinator lifecycle

lucas-albers-lz4:
## Summary

- extract poll request ownership into a dependency-injected coordinator
- preserve one active request plus one coalesced follow-up across visibility and cadence changes
- keep fetch, normalization, buffering, rendering, table pause/resume, hostname handling, and DOM work in the view
- add coordinator behavior coverage, loader/smoke wiring, and architecture documentation

Refs #352

## Validation

- `npm run lint:js`
- `npm run lint:format`
- `git diff --check`
- `./scripts/fwlive-test.sh`

The full suite passes. The existing environment skips only checks requiring unavailable host/device tooling, including real `jshn` poll-cap coverage and fresh POT parity scanning.

Draft for review; do not merge yet.

## Summary by CodeRabbit

* **Enhancements**
  * Improved polling reliability by coordinating overlapping refresh requests and preventing duplicate work.
  * Polling now responds more consistently to visibility changes and cadence adjustments.
  * Leaving the view cleanly settles active requests and prevents departed views from applying late results.
* **Documentation**
  * Updated architecture and package documentation to describe polling coordination and lifecycle behavior.
* **Tests**
  * Expanded coverage for polling, visibility changes, backoff behavior, disposal, request failures, and late responses.


docs/developer/architecture.md
```diff
@@ -33,7 +33,8 @@ flowchart TB
 
 | Module | Location | Responsibility |
 |--------|----------|----------------|
-| **View shell** | `htdocs/.../view/status/fwlive.js` | Layout, 1s poll, pause/limit, DOM, click-to-filter |
+| **View shell** | `htdocs/.../view/status/fwlive.js` | Layout, fetch/render handlers, pause/limit, DOM, click-to-filter |
+| **Poll coordinator** | `htdocs/.../fwlive/poll-coordinator.js` | One active request, one coalesced follow-up, epoch/visibility lifecycle, cadence, and pagehide disposal |
```
lucas-albers-lz4: Two doc mismatches on this new row:

1. **pagehide** is still bound in `view/status/fwlive.js` `load()`, which then calls `dispose()`. The coordinator owns disposal, not the `pagehide` listener. Please either move the listener into the module or drop “pagehide disposal” from this sentence.

2. **`openwrt-feed/luci-app-fwlive/README.md`** still lists every `resources/fwlive/*.js` module except this one. That table is the shipped-module list; it should gain a `poll-coordinator.js` row in this PR.

openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js
```diff
@@ -1050,36 +1053,61 @@ return view.extend({
...
+			}.bind(this),
+			onStateChange: function (state) {
+				this.pollDataInFlight = state.inFlight;
+				this.pollRequestPromise = state.promise;
+				this.pollRequestQueued = state.queued;
+				this.pollRequestWaiters = state.waiters;
+				this.pollRequestQueuedWaiters = state.queuedWaiters;
+				this.coordinatorDisposed = state.disposed;
+				this.pollFn = state.pollFn;
+				this.pollCadenceSec = state.cadenceSec;
```
lucas-albers-lz4: Request start is in the coordinator, but these view fields remain a mirrored copy (`inFlight`, waiters, `pollFn`, cadence, disposed). Tests still assert on the view side, which is why this exists, but it is now easy to “fix” a test by writing `this.pollRequestQueued` and miss the coordinator.

If the mirror is staying for this extract, a one-line comment here that the coordinator is SoT and these are projections would help. Longer-term the view tests should read `this.pollCoordinator.getState()` and these fields can go.

openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js
```diff
@@ -1050,36 +1053,61 @@ return view.extend({
...
-		this.pollCadenceSec = next;
-		try {
-			poll.add(this.pollFn, next);
-		} catch (e) {
-			/* poll gone */
-		}
+		const coordinator = this.ensurePollCoordinator();
+		if (!coordinator.getState().pollFn && this.pollFn)
+			coordinator.adoptPollFunction(this.pollFn);
+		coordinator.setCadence(sec);
```
lucas-albers-lz4: `adoptPollFunction` is only needed because some tests assign `v.pollFn = v.pollData.bind(v)` before `setPollCadence`. Production `load()` goes through `startPolling()`, which installs `requestPoll` itself.

If a future caller sets `this.pollFn` to something else before the first cadence change, LuCI will tick that function instead of the coordinator. Either drop the shim and update the tests to `startPolling()`, or gate it so production cannot adopt a non-coordinator `pollFn`.

openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js
```diff
@@ -1110,81 +1138,30 @@ return view.extend({
...
 	},
 
 	onVisibilityChange() {
-		if (this.isTabHidden()) {
-			this.stopPollingForHidden();
-			this.bumpPollEpoch();
-			return;
-		}
-		this.resumePollingAfterVisible();
+		this.ensurePollCoordinator().onVisibilityChange();
```
lucas-albers-lz4: `resumePollingAfterVisible()` now aliases `onVisibilityChange()`. Calling it while the tab is hidden takes the **hide** path (remove poll, bump epoch, no `onVisible` / RTT reset), unlike the old method which always bumped, reset RTT, and `requestPoll()`’d.

`testEpochDiscardsStale` calls this while visible, so it still passes. If the wrapper is only for that test, consider calling `requestPoll()` after an explicit visible transition instead of overloading the hide/show handler.

openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/fwlive/poll-coordinator.js
```diff
@@ -0,0 +1,260 @@
...
+			return promise;
+		}
+
+		if (state.queued) {
+			queuedWaiters.push(waiter);
+			const queued = queuedWaiters;
+			queuedWaiters = [];
+			state.queued = false;
+			start(queued);
+			return promise;
```
lucas-albers-lz4: This is the Limit-while-hidden catch-up: `finish()` while hidden leaves `queued` set with `currentRun == null`, then show calls `requestPoll()` and starts here.

#352 acceptance asked the **coordinator** tests to cover combined Limit + visibility. `testVisibilityCatchup` queues only after show, while the in-flight request is still running (`currentRun` set), so it never enters this branch. `tests/fwlive-view-fetch-budget.test.js` covers it at the view. Please add a module test: `requestPoll` → second `requestPoll` (queue) → hide → release → assert no third run → show → one catch-up.

tests/fwlive-poll-coordinator.test.js
```diff
@@ -0,0 +1,158 @@
...
+	assert.strictEqual(h.runs.length, 1, 'disposed requests must not start work');
+	console.log('poll coordinator: disposal settles and discards OK');
+}
+
+(async function main() {
+	await testCoalescing();
+	await testVisibilityCatchup();
+	await testCadenceRegistration();
+	await testDisposeSettlesAndDiscards();
+	console.log('poll coordinator tests passed');
```
lucas-albers-lz4, line 309: Coalescing, hide/show, cadence, and dispose are here; combined Limit + visibility is not. The missing case is the `requestPoll` branch where `queued` is true and `currentRun` is null after a hidden finish (see coordinator `requestPoll`).

scripts/fwlive-test.sh
```diff
@@ -137,6 +137,8 @@ echo "== fwlive view poll contract (#233 / #240) =="
 
 echo "== fwlive view poll guard (#240) =="
 "$NODE" tests/fwlive-view-poll-guard.test.js
+echo "== fwlive poll coordinator (#352) =="
+"$NODE" tests/fwlive-poll-coordinator.test.js
 "$NODE" tests/fwlive-view-layer2-backoff.test.js
```
lucas-albers-lz4, line 143: This banner now also covers `fwlive-view-layer2-backoff.test.js` (it never had its own header on master). Either give layer2 its own `echo` or keep coordinator + layer2 grouped under a name that says both.

lucas-albers-lz4 (commented):
## Review (grok)

Faithful extract of the #306/#351 poll lifecycle. I did not find a production overlap/discard/dispose break; existing layer2 and fetch-budget tests still encode those contracts.

Inline notes are docs/test-coverage and a dual-state footgun, not merge-blocking functional defects. Still draft: CodeRabbit skipped this PR, and the repo cycle still wants Bugbot before marking ready.

CI is green on this head.

lucas-albers-lz4:
# Multi-Model Review — lucas-albers-lz4/fwlive PR #353

> **Head SHA:** `9bf2db2792e90b6326d47dbb96ebf6b38d8d6be3`  •  **Reviewers:** muse-spark-1.3-contributor, glm-5.3-flash, deepseek-v4.1-flash  •  **Synthesized by:** deepseek-v4-pro

## TL;DR
The PR extracts the view's single-flight poll lifecycle (one active request + one coalesced follow-up, epoch bumps, visibility/cadence, dispose) into a dependency-injected `fwlive/poll-coordinator.js`, leaving fetch/render/DOM in `fwlive.js` and mirroring coordinator state back via `onStateChange` so existing view tests stay green. All three reviewers agree the core state machine is faithful and well-tested, with no critical bugs. The consensus concern is **lifecycle idempotency and cadence plumbing at the view seam** — several independent Major/minor flags point to the same few gaps (unconditional `poll.add`, terminal `dispose()` never clearing the cached coordinator, a hardcoded `1` cadence fallback, and a raw `pollCadenceSec` write that `publish()` clobbers). Merge is close, but the author should close the lifecycle guards and add the missing coordinator tests first.

### 🔴 Critical
None.

### 🟡 Major

- **Finding:** Cadence value is hardcoded/duplicated outside the injected module and a raw field write bypasses coordinator state.
- **File:** `openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/fwlive/poll-coordinator.js:28,146,166` and `view/status/fwlive.js:2081`
- **Reviewers who flagged:** A (Major), B (Minor), C (Minor) — the single most-flagged issue (3/3).
- **Problem:** Three independent seams all drift from `constants.POLL_CADENCE_FAST_S`: the `initialCadence`/`setCadence` fallback is the literal `1`, and `load()` assigns `this.pollCadenceSec = constants.POLL_CADENCE_FAST_S` directly **after** `bindVisibility()` already created the coordinator with `initialCadence = this.pollCadenceSec` (the view's default `1`). The subsequent `startPolling() → publish() → onStateChange` overwrites that assignment from `state.cadenceSec`, so the raw write is a latent no-op. The mechanism is a real state-consistency bug; whether it *observably* drifts today depends on `POLL_CADENCE_FAST_S === 1` (which I could not verify — the working dir only holds the diff, not `constants.js`).
- **Suggested fix:** Inject `defaultCadence`/`minCadence` (or `'require constants'`) into the coordinator; route startup cadence through `setPollCadence(FAST)` rather than a raw field write.
- **Confidence:** high (mechanism) / medium (live-vs-latent impact).

- **Finding:** `startPolling()` re-registers the poll fn unconditionally and `dispose()` is terminal with no way to clear the cached coordinator.
- **File:** `openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/fwlive/poll-coordinator.js:218-226` and `view/status/fwlive.js:1168-1170`
- **Reviewers who flagged:** A (Major), C (Major, low).
- **Problem:** `startPolling()` dropped the old `if (!pollFn)` guard, so a second `load()` on the same view instance calls `poll.add` again (double tick rate). Relatedly, `dispose()` sets `state.disposed = true` permanently and `startPolling()` then silently no-ops, while the view's cached `this.pollCoordinator` is never cleared — leaving polling dead with no error on bfcache/SPA re-entry.
- **Suggested fix:** Guard re-entry in `startPolling()` (or remove-then-add), and have the view clear/recreate `this.pollCoordinator` when `getState().disposed` (or on a fresh `load()`).
- **Confidence:** medium (double-register path unproven; LuCI `poll.add` dedupe unverified by any reviewer).

- **Finding:** `adoptPollFunction` can install legacy `pollData` as the tick, bypassing the coalescing invariant.
- **File:** `view/status/fwlive.js:1106-1111`
- **Reviewers who flagged:** A (Major).
- **Problem:** If `this.pollFn` is ever set to `pollData` (legacy path) before the coordinator owns a poll fn, `adoptPollFunction` installs it, so the LuCI tick fires `pollData` directly and skips the coordinator's one-active-request machinery. In the current rewrite no production path sets `pollFn = pollData`, so this is a latent/back-compat hazard rather than a live bug.
- **Suggested fix:** Drop the adopt path, or assert the adopted fn is `requestPoll` (fail loudly otherwise).
- **Confidence:** medium.

### 🟢 Minor

- **Finding:** Disposed-guard asymmetry in `bind()` and the hidden branch of `onVisibilityChange()`.
- **File:** `poll-coordinator.js:187-206`
- **Reviewers who flagged:** B, C.
- **Problem:** The visible branch checks `state.disposed` (line 174), but the hidden branch (removePoll + bumpEpoch + publish) and `bind()` do not. A `visibilitychange` after `dispose()` mutates a dead coordinator and fires `onEpochChange` into the view.
- **Suggested fix:** Early-return in `bind()` and the hidden branch when `state.disposed`.
- **Confidence:** medium.

- **Finding:** Test coverage gaps on the new public surface.
- **File:** `tests/fwlive-poll-coordinator.test.js`
- **Reviewers who flagged:** A (Minor), B (Nit), C (Minor).
- **Problem:** No coverage for a rejected/throwing `run` (the new `Promise.reject` path at `poll-coordinator.js:102`), sync-throwing `run`, hidden `requestPoll` no-op, double `startPolling`, restart-after-dispose, `stopPolling`, `adoptPollFunction`, or `unbind`.
- **Suggested fix:** Add four small cases: rejected run settles waiters + clears `inFlight`; `stopPolling` removes without epoch bump; `adoptPollFunction` publishes `pollFn`; restart-after-dispose is a no-op.
- **Confidence:** high.

- **Finding:** Dead fields and methods left at the seam.
- **File:** `view/status/fwlive.js:1152-1165,2020-2022,1140-1142`
- **Reviewers who flagged:** A, C.
- **Problem:** `visibilityHandler` is now always null and `visibilityBound` is set unconditionally, so the flag no longer reflects registration; `pollData`/`stopPollingForHidden` are now production-dead (tests only), and `pollRequestPromise`/`pollRequestWaiters`/`pollDataInFlight` are coordinator-owned mirrors that can drift.
- **Suggested fix:** Remove dead fields or return `bind()`'s real result; mark/drop `pollData` and route its tests through the coordinator API.
- **Confidence:** high.

- **Finding:** `disposeCoordinator()` uses `ensurePollCoordinator()`, creating a coordinator just to dispose it.
- **File:** `view/status/fwlive.js:1163-1165`
- **Reviewers who flagged:** A.
- **Problem:** On a never-started view this allocates a coordinator and fires an epoch bump/publish for no reason.
- **Suggested fix:** `if (this.pollCoordinator) this.pollCoordinator.dispose();` (also subsumes the Major "clear cached coordinator" fix).
- **Confidence:** high.

- **Finding:** `start()` invokes `runRequest(epoch)` before assigning `currentRun`, allowing a synchronous re-entrant `requestPoll()` to start a second overlapping request.
- **File:** `poll-coordinator.js:95-106`
- **Reviewers who flagged:** C (Major, low).
- **Problem:** `state.inFlight` is set before `runRequest`, but `currentRun` is assigned after; a synchronous re-entrant caller sees `currentRun === null` and `queued === false`, so it starts another run. **Note:** the old `startPollRequest()` had the identical ordering (`runPollRequest` before `pollRequestPromise`), so this is pre-existing behavior, not a regression — hence my downgrade to Minor.
- **Suggested fix:** Assign `currentRun`/`inFlight` before invoking `runRequest`, then publish.
- **Confidence:** low.

### 💡 Nits

- **Finding:** Manual index loop in `settle()` instead of `forEach`/`for...of`. — **File:** `poll-coordinator.js:7-9` — **Reviewers who flagged:** A — Confidence: high.
- **Finding:** `publish()` slices both waiter arrays on every state change (1s hot path allocation). — **File:** `poll-coordinator.js:39-52` — **Reviewers who flagged:** C — Confidence: high.
- **Finding:** `getState()` omits `promise`/`waiters` that `publish()` exposes — asymmetric public shape. — **File:** `poll-coordinator.js:237-246` — **Reviewers who flagged:** C — Confidence: high.
- **Finding:** Docs say the coordinator owns "pagehide disposal", but pagehide binding stays in `fwlive.js`; the coordinator only performs teardown when asked. — **File:** `docs/developer/architecture.md:35` — **Reviewers who flagged:** B — Confidence: high.
- **Finding:** Stale "pagehide disposal" comment still in the view. — **File:** `view/status/fwlive.js:2081` — **Reviewers who flagged:** C — Confidence: medium.

## Reviewer agreement matrix

|---|---|---|---|---|
| Hardcoded `1` cadence fallback (3 places) | ✓ | ✓ | ✓ | A=Major, B/C=Minor |
| `load()` raw `pollCadenceSec` write clobbered by publish | ✓ | — | ✓ | Same cadence-plumbing root cause |
| `startPolling()` unconditional add / double register | ✓ | — | ✓ | A and C both flag the lifecycle seam |
| `dispose()` terminal + cached coordinator never cleared | — | — | ✓ | Only C; related to A's dispose-via-ensure Minor |
| `adoptPollFunction` bypasses coalescing | ✓ | — | — | Only A |
| Disposed-guard missing in `bind()` + hidden visibility branch | — | ✓ | ✓ | B and C agree; concrete inconsistency |
| `start()` reentrancy (runRequest before currentRun) | — | — | ✓ | C=Major low; I downgrade (pre-existing) |
| Dead fields (`visibilityBound`/`visibilityHandler`) | ✓ | — | ✓ | |
| `disposeCoordinator` creates-then-disposes | ✓ | — | — | |
| Test coverage gaps | ✓ | ✓ | ✓ | Different specific gaps, same theme |

- **Severity of the cadence literal:** A called the hardcoded `1` fallback Major (high); B and C called it Minor (B "silent divergence", C "duplicating constant"). My take: the *mechanism* (raw write + clobber) is a genuine state-consistency bug, but whether it causes *observable* drift depends on `POLL_CADENCE_FAST_S === 1`, which none of us could verify from the diff alone — so I kept it Major but flagged impact as medium. Human call: check `constants.js`.
- **Double-`startPolling` realness:** A treats it as a medium Major; B and C raise it only obliquely (B via "test-only double-add", C via "restart-after-dispose"). All three acknowledge the re-entry path is unproven in production. My take: fold both A and C's lifecycle points into one Major and let the author confirm whether LuCI `poll.add` dedupes identical fns.
- No disagreement on the core state machine — all three independently verified coalescing, visibility catch-up, and dispose-identity guards as correct.

- **The reentrancy ordering is not a regression.** C's Major on `start()` is pre-existing: the old `startPollRequest()` also called `runPollRequest(epoch)` before assigning `pollRequestPromise`. Worth recording as a non-regression so it isn't treated as a blocker (downgraded to Minor above).
- **`setPollCadence`'s `getState().pollFn` read is safe today but fragile.** The view syncs `this.pollFn` via `onPollFunction`/`onStateChange` to the coordinator's `requestPoll`; the adopt fallback only fires if that sync hasn't happened. No reviewer tied the adopt hazard to the ordering of the sync — the fix for the `adoptPollFunction` Major should also verify the sync always precedes any `setPollCadence`.
- **`pagehide` listener is added but never removed** and `pagehideBound` is only ever set once; combined with the terminal `dispose()`, a full SPA re-entry that reuses the view instance would leave the old listener and a dead coordinator. Same root as the "clear cached coordinator" Major, worth one explicit line in the fix.

- [ ] Inject cadence default (or `'require constants'`) and route `load()` cadence through `setPollCadence(FAST)` instead of a raw field write.
- [ ] Guard `startPolling()` against re-registration (remove-then-add), and clear/recreate `this.pollCoordinator` when disposed or on a fresh `load()`.
- [ ] Add `state.disposed` guards to `bind()` and the hidden branch of `onVisibilityChange()`.
- [ ] Drop or assert-guard `adoptPollFunction`; remove dead `visibilityBound`/`visibilityHandler`/`pollData` fields (or mark them).
- [ ] Add the four coordinator tests (rejected run, stopPolling, adopt, restart-after-dispose) plus a sync-throw and hidden-`requestPoll` no-op case.
- [ ] Fix the `architecture.md` "pagehide disposal" wording and the stale view comment.

- Reviewer A (muse-spark-1.3-contributor): Most aggressive severity (4 Major); targeted the view seam and lifecycle idempotency precisely, but did not run tests and hedged on LuCI `poll` semantics/instance reuse.
- Reviewer B (glm-5.3-flash): Most thorough verification (ran the unit test, fetch-budget, smoke, eslint, full `fwlive-test.sh`); conservative on severity (no Major) but flagged the concrete disposed-guard asymmetry and a docs nit.
- Reviewer C (deepseek-v4.1-flash): Ran the relevant node suites green and correctly identified the reentrancy + terminal-dispose issues, but its "environment skips" verification claim conflicts with B's (C says the script aborts hard at the jshn gate before the new test; B says only that pre-existing failure). Both agree the new/affected node tests pass when run individually.

## Process notes
- Wall time: ~30 seconds (synthesis only; three prior reviewer passes).
- Reviewer failures: none (A, B, C all produced reports).
- Diff was truncated: no (833 lines, complete).
- Total cost: low — single small diff, three focused reviews, one consolidation.

docs/developer/architecture.md
lucas-albers-lz4: FOLD: architecture now says the view owns `pagehide` and calls `disposeView()`; package README lists `poll-coordinator.js` without an upstream-relative link.

openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js
lucas-albers-lz4: FOLD: `onStateChange` mirroring is gone. Coordinator state is only `getState()`; the view reads epoch via `currentPollEpoch()`.

openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js
lucas-albers-lz4: FOLD: `adoptPollFunction` is gone. Tests call `startPolling()`; `poll.add` always registers `requestPoll`.

openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js
lucas-albers-lz4: FOLD: `resumePollingAfterVisible` / `onVisibilityChange` wrappers are gone. Visibility is subscribed in `startPolling()`; tests drive hide/show through the document adapter.

openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/fwlive/poll-coordinator.js
lucas-albers-lz4: FOLD: `testQueuedCompletionWhileHidden` covers queued Limit intent surviving hidden completion, then one visible catch-up that reads the latest budget. Shared `pending` batch replaces the waiter-array branch.

tests/fwlive-poll-coordinator.test.js
lucas-albers-lz4: FOLD: coordinator tests now include Limit-while-hidden, idempotent registration, runner failure recovery, and sync reentry/disposal.

scripts/fwlive-test.sh
lucas-albers-lz4: FOLD: layer2 has its own `echo` header again.

Issue #190, opened 2026-08-21
Security: is_resolvable_address admits hostname-shaped tokens — read-ACL sessions can trigger arbitrary upstream DNS queries


lucas-albers-lz4:
## Summary

`is_resolvable_address` (`openwrt-feed/luci-app-fwlive/root/usr/libexec/rpcd/fwlive:306-316`) claims in its comment to "Accept only IPv4/IPv6-shaped tokens … reject shell metachar / hostnames", but the implementation is a character-class check:

```sh
case "$addr" in
    *[!0-9a-fA-F:.]*) return 1 ;;
    *.*.*.*|*:* ) ;;
    *) return 1 ;;
esac
```

Any token composed solely of `[0-9a-fA-F:.]` with four+ dotted groups (or a colon) passes — including strings that are **not** IP addresses: `ab.cd.ef.01`, `dead.beef.cafe.baad`, `a.b.c.d.e.f`. The sink `getent hosts "$ip"` (`rpcd/fwlive:293-304`) falls back to a nameserver lookup for such tokens.

The selftest (`rpcd/fwlive:389-409`) covers spaces, metacharacters, `example.com`, a real IPv4 and an IPv6 address — but no dotted-hex hostname-shaped case, so the claimed "reject hostnames" behavior is untested for this class.

## Impact

Any ubus session holding the read ACL (which includes `resolve`, `acl.d/luci-app-fwlive.json:6`) can call `ubus call fwlive.resolve '{"addresses":["aa.bb.11.22"]}'` and make the router issue an attacker-chosen upstream DNS query — a covert channel / DNS-probe primitive from a low-privilege authenticated session. The returned (attacker-influenced) name is displayed in the LuCI table; no shell injection (argv-only) and no XSS (text-node DOM) were found in the path.

## Notes

- Severity: Low
- Found by read-only security review (zen x-preview-f-free @ max reasoning, 2026-08-21); verified against `master` @ `3f63bb9`
- Facts only — fix/plan handled separately per repo convention

lucas-albers-lz4:
## Fix shipped — PR #192 (open, awaiting merge approval)

[...]

PR #358, opened 2026-09-17
refactor: keep feed comments luci-shaped

### LINKED ISSUE #336: Decide whether fwlive comments should drop tracker ids to match luci
## Decision (not this luci wave)

After [openwrt/luci#8992](https://github.com/openwrt/luci/pull/8992), shipped LuCI comments must not use `#N` tracker ids: GitHub auto-links them to the **luci** issue tracker (BKPepe on `buffer.js` `#43` → [openwrt/luci#43](https://github.com/openwrt/luci/issues/43), which is unrelated).

For the current recut we only drop the `#43` comments that maintainers actually hit, and we do **not** rewrite the rest of the package comments.

This issue is: should **fwlive** source comments (`openwrt-feed/…`) later match that luci style (behavior-only, no `#N`), or keep tracker ids here because they are useful while developing?

Keep `#N` in tests, `CHANGELOG.md`, and commit messages either way.

## Options

1. **Keep `#N` in fwlive comments.** Luci copies stay clean via the cut (or a later cut-script sanitizer). Dual comment style.
2. **Match luci in fwlive.** Strip `(#N)` / `issue #N` from package-tree comments in this repo too. One style; lose in-file jumps to issues.

Do not use a global `#\d+` replace (`css.js` / `tint.js` hex colors).

Parent: [#209](https://github.com/lucas-albers-lz4/fwlive/issues/209).

lucas-albers-lz4:
## Summary

- Remove local tracker IDs from shipped feed comments while preserving behavior-focused explanations.
- Update the classifier generator and regenerate its shipped shell wrapper.
- Add a source gate matching the upstream-cut comment policy.

## Verification

- ./scripts/fwlive-test.sh
- Source and cut comment gates pass.

Closes #336

## Summary by CodeRabbit

- **Documentation**
  - Clarified inline guidance for firewall event handling, adaptive polling, health monitoring, logging, temporary-file security, and address resolution.
  - Removed outdated issue-tracking references from technical comments.

- **Tests**
  - Improved comment-policy validation across source and generated outputs.
  - Added broader checks for prohibited repository references and tracker IDs.
  - Improved discovery of the internationalization scanner in alternate development environments.


openwrt-feed/luci-app-fwlive/root/usr/libexec/rpcd/fwlive
```diff
@@ -545,7 +545,7 @@ is_resolvable_address() {
 	# Reject anything outside the address alphabet BEFORE awk, so embedded
 	# newlines/whitespace (which awk would treat as record separators) can
 	# never let a valid first record mask trailing garbage that still reaches
-	# nslookup (#190, CodeRabbit/luna fold): nslookup only ever sees a single
+	# nslookup only ever sees a single
```
lucas-albers-lz4: Nit: removing `(#190, CodeRabbit/luna fold):` took the clause joiner with it. Lines 545-549 now read "...trailing garbage that still reaches / nslookup only ever sees a single / token." Suggest restoring the break, e.g. `# nslookup: nslookup only ever sees a single` or an em-dash, so the two clauses don't fuse.

tests/fwlive-upstream-cut.test.sh
```diff
@@ -82,6 +82,38 @@ grep -q '^## Dependencies$' "$CUT_WORK/cut/README.md" \
...
+# Keep the feed source in the same comment shape as the luci-shaped cut. The
+# generated awk classifier is exempt because it is regenerated from the core
+# source; css.js and tint.js are exempt because their comment-like content can
+# contain CSS color values. Tests, changelog, and repo docs keep tracker ids.
+python3 - "$ROOT/openwrt-feed" <<'PY' || die "feed comments contain tracker ids"
+import re, sys
+from pathlib import Path
+
+root = Path(sys.argv[1])
+pat = re.compile(r'(?i)(?:issue\s+|Grok\s+)?#\d{2,}')
```
lucas-albers-lz4: Question, non-blocking: `\d{2,}` lets single-digit `#N` refs (e.g. `#5`) through the gate. That also dodges false positives like "option #1", so it's probably deliberate — but worth one line in the comment above stating single-digit refs are intentionally out of scope, so a future reader doesn't "tighten" it to `\d+` and break the gate on prose.

lucas-albers-lz4 (commented):
Review of 20b9a49 (draft): direction matches issue #336 option 2, gate re-run locally on the PR head reports 0 hits, and the regenerated shell classifier matches its generator. Two minor nits inline, nothing blocking.

lucas-albers-lz4:
## Review: refactor to keep feed comments luci-shaped

**Verdict: no blocking issues** (draft PR, two minor nits as inline comments).

### What I verified
- Re-ran the new feed gate logic against the PR head (`20b9a49`): **0 hits** — all `#NN` tracker ids are gone from `openwrt-feed` comments outside the declared exemptions (`css.js`, `tint.js`, `fwlive-is-firewall-event.awk`, `.pot`).
- Generator parity holds: `scripts/gen-shell-classifier.js` comment change and the regenerated `fwlive-is-firewall-event.sh` match exactly; no hand-edit of the generated file.
- The stripped comments all retain their behavioral rationale (fail-closed, first-wins, strict-mode ownership); only the tracker pointers were removed.
- Scope looks right: `CHANGELOG.md` (repo root) and `tests/` sit outside `openwrt-feed`, so the "keep `#N` in tests/changelog/commit messages" carve-out from #336 holds automatically.
- CI is green across the board (fwlive-test, CodeQL, zizmor, z3-verify) and the branch is mergeable. Body uses same-repo `Closes #336`, which is fine.

### Inline nits (minor)
1. `rpcd/fwlive:548` — fused sentence after the parenthetical (with colon) was removed.
2. `tests/fwlive-upstream-cut.test.sh:94` — consider documenting that the `\d{2,}` threshold intentionally excludes single-digit refs.

### Before merge
- This is still a **draft**, so CodeRabbit won't auto-review it. Marking it Ready makes it eligible for automatic review per the repo's review setup — suggest doing that and batching any follow-ups rather than merging on CI-green alone.

tests/fwlive-upstream-cut.test.sh
```diff
@@ -82,6 +82,38 @@ grep -q '^## Dependencies$' "$CUT_WORK/cut/README.md" \
...
+skip_names = {'css.js', 'tint.js', 'fwlive-is-firewall-event.awk'}
```
lucas-albers-lz4: The skip list is broader than the scanner needs, and it fights the invariant in the comment above ("same comment shape as the luci-shaped cut").

The scanner only looks at lines whose stripped form starts with `#` / `//` / `/*` / `* `, and it already ignores hex-looking `#` tokens. That already covers:
- `css.js` — colors live in `styleText: "..."` (the line does not start with a comment prefix)
- `tint.js` — colors live in `var … = '#46a546'` assignments, not comments

Skipping those two files does not avoid a current false positive. It does mean a later `// (#43)` in `css.js` or `tint.js` would not fail this gate, which is the original luci auto-link footgun.

`fwlive-is-firewall-event.awk` is skipped here but still scanned (and sanitized) on the cut side, so source and cut can diverge for a shipped asset. If the goal is one comment shape, drop the awk exemption and keep the generator clean instead.

These two Python blocks will also drift independently. Worth extracting one helper used by both the source and cut checks.

lucas-albers-lz4 (commented):
Comment-policy change matches #336: tracker ids leave shipped feed comments, behavior notes stay, and the generator stays in sync with the wrapper. JS/CSS in the feed were already clean; the remaining `#N` notes were in the ash plugin and the generated classifier header. CI is green.

One leftover strip broke a sentence in `is_resolvable_address`. The new source gate also skips `css.js` / `tint.js` / the awk classifier even though the scanner already ignores non-comment lines and hex tokens — that skip is the hole that would let a `#43`-style comment back into the files this policy exists to protect.

openwrt-feed/luci-app-fwlive/root/usr/libexec/rpcd/fwlive
lucas-albers-lz4: Addressed in 4a7a74f: the comment now ends the first clause after `nslookup` and states that it must receive a single token.

tests/fwlive-upstream-cut.test.sh
lucas-albers-lz4: Addressed in 4a7a74f: the scanner now documents that the two-digit threshold intentionally leaves single-digit #N prose references out of scope.

tests/fwlive-upstream-cut.test.sh
lucas-albers-lz4: Addressed in 4a7a74f and 0b9c891: the source and cut checks now share one scanner, the file exemptions are gone, the cut-only repository-name check is parameterized, and the sibling LuCI checkout is discovered automatically when available.

PR #359, opened 2026-09-17
Refactor render scheduling ownership

lucas-albers-lz4:
## Problem

Render frame coordination and render token-bucket state were embedded in the LuCI view, making epoch handling, forced repaint intent, and DOM ownership harder to follow.

## Result

- Extract render scheduling into a DOM-free `render-scheduler.js` module.
- Keep epoch-safe frame coalescing, stale-frame requeue, token budgeting, and disposal in one owner.
- Preserve forced repaint intent across refresh cleanup and hostname-resolution timing.
- Keep DOM construction, status banners, and table state in the view.
- Add scheduler and lifecycle regression coverage plus architecture documentation.

## Validation

- `./scripts/fwlive-test.sh`
- `./scripts/validate-baseline.sh`
- `npm run test:view`
- Link checking and `git diff --check`
- Commit secret scan and Z3 verification

Refs #340

## Summary by CodeRabbit

- **New Features**
  - Added smarter live-view rendering with frame coalescing and render-rate limiting.
  - Added forced repaint support for row-limit changes and other urgent updates.
  - Added safeguards to discard stale queued paints and cancel pending work during disposal.
  - Added synchronous rendering fallback when animation-frame support is unavailable.

- **Documentation**
  - Documented the new rendering behavior and module in developer and package documentation.

- **Tests**
  - Added coverage for throttling, forced renders, stale frames, disposal, and row-limit refresh behavior.


lucas-albers-lz4 (commented):
Review of 274e213e (draft): extraction matches the #340 ownership map (scheduler owns frame intent + token bucket, view keeps DOM), epoch/force/budget semantics trace equivalent to the removed view code, and all JS suites I ran on the PR head pass. One convention nit inline, nothing blocking.

tests/fwlive-render-scheduler.test.js
```diff
@@ -0,0 +1,202 @@
...
+			assert.ok(frames.length, 'expected a pending frame');
+			frames.shift()();
+		}
+	};
+}
+
+/* The policy consumes the last painted identity, not merely an allowed paint. */
+{
+	const s = makeScheduler();
+	assert.equal(s.shouldRender(rows(['a']), false, 99), true, 'count changes cost one');
```
lucas-albers-lz4: Convention nit, non-blocking: this file uses `assert.equal` throughout (loose `==`), while the sibling suites (`layer2-backoff`, `render-policy`) use `strictEqual`/`deepStrictEqual`. No live bug — `shouldRender`/`isFloodSuppressed` always return real booleans — but strict assertions would match repo convention and guard against a future non-boolean return silently passing.

lucas-albers-lz4:
## Review: isolate render scheduling state

**Verdict: no blocking issues** (draft PR, one convention nit as an inline comment).

### What I verified (on PR head `274e213e`)
- **Tests, all passing:** `fwlive-render-scheduler`, `fwlive-render-policy`, `fwlive-view-layer2-backoff` (incl. the two new Limit/force tests), `view-poll-error`, `view-poll-guard`, `poll-coordinator`, `view-fetch-budget`, `modules-smoke`, plus `npm run test:view`.
- **Behavior preservation** (old view code vs new scheduler, traced by hand):
  - Epoch/stale-frame handling is equivalent — stale frames drop without inventing work, and a current request queued behind stale work still gets a fresh frame with force retained.
  - Force reservation lifecycle matches: queued force survives refresh cleanup, per-reservation cancel can't erase queued or newer force, force expires after painting.
  - Token bucket matches: first-call refill, clock-rollback (`max(0, …)`), capacity cap, forced paints bypassing without clearing the warning, identity committed only via `markRendered` after paint.
  - `resetBudget()` setting `bucketMs = now()` is a no-op difference vs the old code (bucket is capped at capacity either way).
- **No leftovers:** no remaining references to the removed view fields (`renderRaf`, `renderBucket`, `pendingForceRender`, …) outside `render-policy`'s own pure API params; `displayRowCap` stays on the view and stays covered by layer2 tests; the removed `renderBudgetCost` delegation assertions are covered by direct `shouldRender` batch-cost tests. New module follows the sibling `baseclass` pattern.
- **Hygiene:** `git diff --check` clean on the PR commit; secret scan shows only benign hits (doc text about scan false positives, a test locator filling an empty password field); `CHANGELOG.md` untouched, consistent with a pure refactor.
- CI is green across the board and the branch is mergeable. Body uses `Refs #340`, no cross-repo close trap.

### Notes (not findings)
- `./scripts/fwlive-test.sh` run locally stops at `fwlive-jshn-compat.test.py`, which needs a host `bin/jshn` install missing in this environment. That test file is untouched by this PR, so the failure is environmental, not PR-related — but the full script hasn't gone green here for that reason.
- Still a **draft**, so CodeRabbit won't auto-review it — same as #358, suggest marking Ready and batching follow-ups rather than merging on CI-green alone.

### Inline nit (minor)
1. `tests/fwlive-render-scheduler.test.js:53` — `assert.equal` vs repo `strictEqual` convention.

lucas-albers-lz4 (commented):
Worktree review of `274e213` (`feat/340-state-dom-ownership`).

GitHub will not accept `REQUEST_CHANGES` on an author-owned PR, so this is a comment review. Treat the RAF wrap as a should-fix before merge.

**Verdict:** merge after wrapping the animation-frame adapters. The scheduler extract is the right ownership split, and Limit/hostname force handling is a real fix versus master wiping `pendingForceRender` in `finally`.

## Should-fix

Unbound `requestAnimationFrame` / `cancelAnimationFrame` (inline on `ensureRenderScheduler`). The scheduler invokes them as loose functions in a strict module; Chrome throws `TypeError: Illegal invocation`. Master called them as bare globals. Node tests inject plain functions, so they cannot catch this.

Wrap at the adapter:

```js
requestFrame: typeof requestAnimationFrame === 'function'
	? (cb) => requestAnimationFrame(cb)
	: null,
cancelFrame: typeof cancelAnimationFrame === 'function'
	? (id) => cancelAnimationFrame(id)
	: null,
```

## What looks good

- Scheduler owns RAF coalesce, epoch requeue, token bucket, last painted identity, force reservation, and dispose; the view still builds DOM.
- `forceNextRender()` is reservation-scoped, so Limit cleanup cannot erase force already queued on a frame.
- Frame id `0` and clock `0` use `!== null` (old `if (this.renderRaf)` / `if (!renderBucketMs)` treated `0` as unset).
- Token refill uses `nowMs()` (`performance.now()`), which is better elapsed-time behavior than wall-clock `Date.now()`.
- Scheduler + layer2 tests passed locally; CI on this SHA is green. `qemu-install-fwlive.sh` already copies every `fwlive/*.js`.

## Nits

- `openwrt-feed/luci-app-fwlive/README.md` still omits `render-scheduler.js` (architecture table was updated).
- `architecture.md` pagehide text still names only the poll coordinator; `disposeView()` now also cancels frames and rejects later paints.
- `testLimitRefreshPreservesQueuedForce` schedules force after `finally()`, so it does not actually interleave hostname force with Limit cleanup.
- Missing blank line in the new architecture scheduler paragraph.

openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js
```diff
@@ -1072,6 +1067,21 @@ return view.extend({
...
+		if (this.renderScheduler) return this.renderScheduler;
+		this.renderScheduler = renderScheduler.create({
+			getEpoch: () => this.currentPollEpoch(),
+			render: (force) => this.renderRows(force),
+			renderCost: renderPolicy.renderCost,
+			now: () => this.nowMs(),
+			capacity: constants.RENDER_CAP_PER_SEC,
+			requestFrame:
+				typeof requestAnimationFrame === 'function' ? requestAnimationFrame : null,
+			cancelFrame: typeof cancelAnimationFrame === 'function' ? cancelAnimationFrame : null
```
lucas-albers-lz4: Passing the native `requestAnimationFrame` / `cancelAnimationFrame` functions into the scheduler means they are invoked as loose calls in a `'use strict'` module. In Chrome that is `TypeError: Illegal invocation`.

Master called them as bare globals. Wrap here so `this` stays the window:

```js
requestFrame: typeof requestAnimationFrame === 'function'
	? (cb) => requestAnimationFrame(cb)
	: null,
cancelFrame: typeof cancelAnimationFrame === 'function'
	? (id) => cancelAnimationFrame(id)
	: null,
```

The Node harness injects plain functions, so existing tests will not catch this.

PR #360, opened 2026-09-17
Guard late startup replies after view disposal

lucas-albers-lz4:
## Summary

- Make view disposal terminal for late startup rules/status RPC replies.
- Keep post-disposal `render()` inert and `addFooter()` a no-op.
- Add success and rejection lifecycle coverage for late startup completion.

Refs #340

## Verification

- `npm run test:view`
- `node tests/fwlive-view-fetch-budget.test.js`
- ESLint and Prettier checks
- `./scripts/fwlive-test.sh`
- `./scripts/validate-baseline.sh`

Astra high-depth review completed on the final commit with no actionable findings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented late startup responses and lifecycle callbacks from updating the view after navigating away.
  * Stopped post-disposal polling, backend requests, and UI updates.
  * Ensured delayed or failed status responses no longer restore or modify disposed page state.

* **Documentation**
  * Clarified the page disposal lifecycle and behavior when startup completes after disposal.


openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js
```diff
@@ -1952,6 +1960,7 @@ return view.extend({
 	},
 
 	load() {
+		if (this.viewDisposed) return Promise.resolve();
```
lucas-albers-lz4, line 1965: Question, non-blocking: the `pagehide` listener is registered inside `load()` itself, so `viewDisposed` can only become `true` after `load()` has already run — is this top-of-`load()` guard reachable in practice? Harmless as defense either way; just checking whether it's intentional belt-and-braces or covering a re-invocation path I'm not seeing.

lucas-albers-lz4 (commented):
Review of 07636c6 (draft): disposal boundary matches the #340 shape ownership, guards cover both success and rejection paths of both startup RPCs, and all suites I ran on the PR head pass. One question inline, nothing blocking.

lucas-albers-lz4:
## Review: ignore late startup replies after view disposal

**Verdict: no blocking issues** (draft PR, one question as an inline comment).

### What I verified (on PR head `07636c6`, based on master incl. #358+#359)
- **Tests, all passing:** full `fwlive-view-fetch-budget` suite incl. the two new tests (`pagehide drops late startup UI` / `...failures`), plus `npm run test:view` exit 0 (run with the repo's `node_modules` linked into the worktree).
- **Guard tracing:** `disposeView` sets `viewDisposed` once and it is never reset — correct for terminal disposal (fresh view instance per LuCI navigation; BFCache/persisted pagehide returns before `disposeView`, covered by the existing persisted test). Both `loadRulesMap` and `loadLoggingStatus` guard try and catch paths, so the trailing `updateBackendUi` / toolbar / empty-state / conditional-`renderRows` calls are unreachable when disposed. `render()` returns an inert div, `addFooter()` is a no-op — the new tests prove all of this with sentinel values (`resolveGeneration` pinned, zero UI updates, zero poll calls).
- **Hygiene:** `git diff --check` clean; secret scan clean; ESLint clean on the view file (test file is under the ignore pattern, as on master); Prettier clean on the shipped file — the test-file Prettier warning is pre-existing on master, not introduced here. `CHANGELOG.md` untouched, consistent with a pure fix. Body uses `Refs #340`, no close trap.
- CI is green across the board and the branch is mergeable.

### Notes (not findings)
- Same draft note as #358/#359: CodeRabbit won't auto-review while this is a draft — suggest marking Ready and batching follow-ups rather than merging on CI-green alone.

### Inline question (minor)
1. `fwlive.js:1963` — top-of-`load()` guard looks unreachable since the pagehide listener is registered inside `load()`; confirming whether it's deliberate defense.

lucas-albers-lz4:
## Multi-Model Review: Grok & Gemini 3.8 Flash

### Grok Review
**Verdict:** APPROVE  
**Findings count:** 2

#### Findings
- **Severity:** Nit  
  **Location:** `openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js:1978`  
  **Issue:** After `Promise.all` for rules/status, `load()` still calls `requestPoll()` with no `viewDisposed` check, relying only on the disposed coordinator to no-op.  
  **Suggested fix:** Guard that `.then()` with `if (this.viewDisposed) return;` so startup’s poll chain matches the new `load()` entry check.
- **Severity:** Nit  
  **Location:** `docs/developer/architecture.md:76`  
  **Issue:** Disposal docs still cover epoch, hostname, coordinator, and scheduler, but not the new terminal `viewDisposed` flag or the `render()` / `addFooter()` no-op contract.  
  **Suggested fix:** Add one sentence note that non-persisted `pagehide` is terminal for late startup RPC application and for post-disposal `render()` / `addFooter()`.

#### Summary
The change is small, correctly placed at the async boundaries, and matches the bug: late `fwlive.rules` / `fwlive.logging_status` replies must not mutate state or paint into leftover DOM, and LuCI finishing `render()` / `addFooter()` after pagehide must not revive the view (including resetting `resolveGeneration`). Tests cover both success and rejection late-reply lifecycles and assert no follow-up poll. Merge-ready after human review; the nits are optional follow-ups, not blockers.

---

### Gemini 3.8 Flash Review
**Verdict:** APPROVE  
**Findings count:** 1

#### Findings
- **Severity:** Nit  
  **Location:** `openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/view/status/fwlive.js:1978`  
  **Issue:** In `load()`, `Promise.all([this.loadRulesMap(), this.loadLoggingStatus()]).then(() => this.requestPoll())` invokes `this.requestPoll()` even if the view was disposed during startup; while `pollCoordinator.requestPoll()` already returns `Promise.resolve()` when disposed, an early return in the `.then()` handler avoids unnecessary method dispatch.  
  **Suggested fix:** Check `if (this.viewDisposed) return;` inside the `.then(...)` handler before calling `this.requestPoll()`.

#### Summary
PR #360 cleanly resolves the race condition where unpersisted `pagehide` teardown occurs while asynchronous startup RPCs (`callFwliveRules` and `callFwliveLoggingStatus`) are still in flight. By establishing `this.viewDisposed` as a terminal lifecycle flag and guarding both resolution and error handlers, stale responses cannot mutate state or trigger DOM renders. Post-disposal `render()` returning an inert `cbi-map` wrapper and `addFooter()` being a no-op prevents LuCI from re-initializing generations or attaching stale event listeners. The accompanying test suite in `tests/fwlive-view-fetch-budget.test.js` provides solid assertion coverage for both success and rejection lifecycles.

Issue #371, opened 2026-09-18
defect: read_nf_log_backend treats uppercase NONE as present logger (fail-open)


lucas-albers-lz4:
# Design and fix contract: make nf_log readiness family-aware and reject NONE

This issue is the single source of truth for the nf_log readiness defect found in #370 R1. Luna completed the consolidated review against the shipped shell helper, status JSON, enable gate, LuCI mapping, tests, documentation, and a live host. No additional model review is needed for the current issue state. The historical comments are folded into this body; future review should add only new deltas or corrections.

## Objective

Prevent a logger-less system from being reported as ready merely because the kernel exposes an nf_log sysctl file containing uppercase NONE, while still allowing IPv4-only deployments to enable WAN logging when IPv6 is genuinely unavailable.

The fix must keep one consistent contract across:

- openwrt-feed/luci-app-fwlive/root/usr/libexec/fwlive-logging.sh;
- read_nf_log_backend, the family checks, blocker collection, status JSON, and the enable gate;
- openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/fwlive/logging.js and view/status/fwlive.js;
- tests/fwlive-logging.test.sh and view/module tests;
- the FAQ and detailed nf_log documentation.

## Confirmed defect

The current helper is effectively:

    read_nf_log_backend() {
        path="$1"
        [ -f "$path" ] || return 1
        val=$(cat "$path" 2>/dev/null) || return 1
        [ -n "$val" ] && [ "$val" != 'none' ]
    }

The kernel writes uppercase NONE when no logger module is registered. On the live host used for review, both /proc/sys/net/netfilter/nf_log/2 and /proc/sys/net/netfilter/nf_log/10 read NONE. The case-sensitive comparison therefore succeeds.

Current impact:

- logging_status reports nf_log_ipv4: true and nf_log_ipv6: true;
- no nf_log_*_missing blockers are reported;
- ready can remain true;
- enable_wan_logging passes its pre-check and stages the WAN log bit;
- the user receives an apparently enabled but permanently empty live view.

The FAQ also describes only lowercase "none" and must be aligned with the final sentinel contract.

## Required state model

Readiness has two independent dimensions for each address family:

1. Family availability: is IPv4/IPv6 actually present and required on this device?
2. Logger backend state: if the family is present, does its nf_log entry contain a usable backend?

[...]

PR #372, opened 2026-09-18
test: implement independent #370 coverage gates

lucas-albers-lz4:
## Summary

Implement the independent #370 coverage gates and close the follow-up review findings.

- Make fuzzy PO entries fail the i18n gate.
- Make the LuCI-accurate table XSS oracle include the intentional root tbody clear while rejecting all payload writes.
- Make the package payload inspector portable on macOS/BSD find and sort.
- Support both release package formats: 24.10 IPK and 25.12 APK.
- Align the test-approach and security-review records with #370 decisions.

## Review status

- Luna independent review found no production security regression; its package-ACL, lock-test, and cache-path findings were fixed.
- Grok review findings are addressed through b6ce5fa.
- The R9a gate builds and inspects both a current 24.10 IPK and a current 25.12 APK; the APK payload is extracted with the matching pinned SDK image.
- R1 remains owned by #371; installed-session, renderer follow-up, and 25.12 manual evidence remain separate #370 work.

## Validation

- ./scripts/fwlive-test.sh - passed
- Real 24.10 IPK payload inspection - passed
- Real 25.12 APK payload inspection - passed
- node tests/fwlive-i18n.test.js - passed
- node tests/fwlive-modules-smoke.test.js - passed
- bash scripts/fwlive-shellcheck.sh - passed
- actionlint and commit hooks - passed

Refs #370


lucas-albers-lz4:
Sync note (2026-09-18): #370 is source of truth for the coverage record. PR body carries the L-to-R mapping plus decision sync (R7 union rule, R1 moved to #371, R6 merge, R3 plus R9 added to minimum-ready). No branch change. Evidence snapshot confirmed valid (PR #368 head).

lucas-albers-lz4:
Luna independent review completed for commit f9df258.

- No production security regression found; targeted tests, BusyBox filter coverage, all five real-jshn compatibility pairs, shell syntax, and diff checks passed.
- Luna's packaged-ACL validation finding was fixed by comparing the extracted ACL semantically with the source and rejecting packaged ubus log access.
- The unused lock-test assignment was removed.
- R9a remains explicitly open: the normal host workflow has no current-version ipk build, so the inspector uses a source-shaped fixture; set FWLIVE_REQUIRE_IPK=1 with a real artifact for package/release validation.

The PR remains draft pending the required external checks and the R9a artifact evidence.

tests/fwlive-i18n.test.js
```diff
@@ -166,13 +222,14 @@ function main() {
...
-    const total = missing + empty + formatMismatch;
+    const total = missing + empty + formatMismatch + fuzzy;
     if (total + mismatched === 0)
       console.log('[PASS] %s: %d msgids OK, %d entries total — formats verified ✓', lang, potIds.length, poEntries.length);
     else {
       if (missing) console.error('[FAIL] %s: %d missing msgid(s)', lang, missing);
       if (empty) console.error('[FAIL] %s: %d empty translation(s)', lang, empty);
       if (formatMismatch) console.error('[FAIL] %s: %d format specifier mismatch(es)', lang, formatMismatch);
+      if (fuzzy) console.error('[FAIL] %s: %d fuzzy translation(s)', lang, fuzzy);
       if (missing + empty + formatMismatch > 0) failures++;
```
lucas-albers-lz4: **CONFIRMED — fuzzy translations do not fail the suite.**

`fuzzy` is included in `total` (so this prints `[FAIL]`) but omitted from `failures++`. `main()` then exits 0 unless another category is also dirty.

`testFuzzyParser()` only checks `parsePoFile()` on a temp file. It never drives this counter. Current catalogs have zero fuzzy entries, so CI stays green. A later `msgmerge` that marks strings fuzzy will print FAIL and still exit 0.

Count fuzzy in `failures++` (or `if (total > 0)`). Drive that through the same path as the live `.po` loop, not only the parser helper. That is the G5 oracle: a temp `.po` with one `#, fuzzy` entry must make `node tests/fwlive-i18n.test.js` exit 1.

tests/fwlive-modules-smoke.test.js
```diff
@@ -370,6 +370,35 @@ keyedTable.renderRows(
...
+);
+assert.ok(
+	collectText(sinkBody).indexOf(hostileMessage) >= 0,
+	'hostile log text must remain visible as text'
+);
+function assertNoHtmlWrites(node) {
+	assert.equal(node._innerHTMLWrites.length, 0, 'hostile log render must not write innerHTML');
+	for (const child of node.childNodes || []) assertNoHtmlWrites(child);
+}
+for (const child of sinkBody.childNodes) assertNoHtmlWrites(child);
```
lucas-albers-lz4: **CONFIRMED — XSS oracle never inspects the node that writes `innerHTML`.**

R2’s SoT is the recording `E()` harness: hostile text visible, payload never in an `innerHTML` write. Production `renderAllRows` clears with `host.innerHTML = ''` (`table.js`). This test skips `sinkBody` and only walks children.

The harness does not parse HTML. `host.innerHTML = hostileMessage` becomes a text child with an empty write log. `collectText` still finds the payload, child writes stay 0, and the test passes.

Walk `sinkBody` itself. Allow only empty-string clears. Assert the hostile bytes appear in no `_innerHTMLWrites` entry (including `title` if that path ever becomes a sink).

docs/developer/test-approach.md
```diff
@@ -184,16 +184,82 @@ necessary history.
...
+| Installed seams | L1 session ACL and the selected L6 matrix cells | The QEMU/manual artifact records image and package/feed revision, install format, architecture, backend, session identity, command/output, date, and cleanup; direct root calls do not count. |
+| Follow-up state coverage | The remaining L8 rollback injections and L9 backend/IPv6/degradation/storage/resolve cases | Each case is split into an independent assertion with an owner and revisit trigger; grouping is allowed for implementation, not for hiding an unverified branch. |
+
+The upstream-cut readiness decision should be made from those exit conditions,
+not from the count of L1–L9 tests. At minimum, L1, L2, L4, L5, and every
+security/configuration branch in L8 need executed evidence or an explicit
+residual; L3 needs a documented contract decision; and L6 needs dated evidence
+for every selected seam cell. A manual result becomes `Manually verified` only
+when its artifact is retained with the review record.
```
lucas-albers-lz4: **CONFIRMED — plan still contradicts #370, which this PR exists to encode.**

The PR body already listed the required sync and said it was not applied. This section still:

- asks L3 to choose a contract #370 already fixed as the union rule (zone name `wan`, or network list carries `wan` / `wan6`)
- puts L2 nf_log fixtures in stage 1 after they moved to #371
- splits L5/L9 across stages after R6 merged them
- omits R3 (`applyHash`) and R9 from the minimum-ready set
- keeps L1–L9 numbering after the body was remapped to R1–R9

Renumber to R1–R9 and apply those decisions. Do not claim executed evidence for items that remain undecided or deferred.

docs/developer/security-review.md
```diff
@@ -26,6 +26,8 @@ reopen an accepted residual without new evidence.
 | Data corruption / availability | Log pipeline failure, lock issues, UCI sweep |
 | Supply chain | Unpinned tooling, signing keys, feed trust |
 
+> **2026-09-19 #370 delta:** Host coverage now pins exact rpcd/ACL method parity and read/write isolation, real reply-shape handling, hostile table text under the recording LuCI E harness, fail-closed shell cases, IPv6 PTR parsing, fuzzy PO rejection, and ipk payload layout/modes. The filter accepts an explicit dump directory for host fidelity, while the production rpcd caller passes /tmp; no ACL grant or DOM sink changed.
```
lucas-albers-lz4: **CONFIRMED — ledger overclaims coverage this PR does not enforce.**

“Fuzzy PO rejection” is the broken gate in `fwlive-i18n.test.js` (FAIL log, exit 0). “Real reply-shape handling” is mostly fixture stuffing: `luci-view-boot.js` still has IPv4-only canned rows, no hostile row, and no assertion that backend / adaptive / IPv6 / hostile content is rendered. That is the failure mode item 5 of this PR’s own plan forbids (“not only fields copied into a mock reply”).

Only claim evidence the tests actually enforce. Keep the filter dump-dir note; that seam is real.

tests/fwlive-ipk-payload.test.sh
```diff
@@ -0,0 +1,102 @@
...
+		fwlive-log-filter.sh|fwlive-logging.sh|rpcd/fwlive)
+			test -x "$EXTRACT/usr/libexec/$rel" ||
+				{ echo "packaged executable lost mode: $rel" >&2; exit 1; }
+			;;
+		*)
+			test ! -x "$EXTRACT/usr/libexec/$rel" ||
+				{ echo "sourced helper unexpectedly executable: $rel" >&2; exit 1; }
+			;;
+	esac
+done < <(cd "$PKG_DIR/root/usr/libexec" && find . -type f -printf '%P\n' | sort)
```
lucas-albers-lz4: **CONFIRMED — GNU `find -printf` breaks the documented macOS host path.**

`docs/developer/environment.md` says macOS runs `fwlive-test.sh`. This is the only test using `-printf`. BSD find will fail this new required gate.

Use a portable listing, e.g. `find . -type f | sed 's|^\./||' | sort`. The module loop’s `sort -z` is also GNU-flavored; keep that POSIX too.

R9a’s source-shaped fixture fallback is fine as an explicit residual, but only if this gate actually runs on the host suite.

tests/fixtures/luci-view-boot.js
```diff
@@ -73,9 +82,11 @@
...
 				wan_log_limit: 10,
 				nf_log_ipv4: true,
-				nf_log_ipv6: false,
+				nf_log_ipv6: true,
 				ready: true,
-				blockers: []
+				weak_device: false,
+				blockers: [],
+				warnings: []
 			};
```
lucas-albers-lz4, line 90: R6 is not closed by copying always-present fields into the mock.

This still has IPv4-only `CANNED_LOGS`, no hostile row, and no test that the view consumes `backend`, adaptive metadata, IPv6, or hostile content. `nf_log_ipv6: true` plus empty `warnings` fixes the old impossible `ready && !nf_log_ipv6 && !blockers` state; it does not pin renderer behavior.

Either add the R6 assertions (backend label, adaptive fields, IPv6 row, hostile row under recording `E()`) or drop the “reply-shape handling” claim and leave R6 open.

lucas-albers-lz4 (commented):
Grok review of `f9df258`. Cannot set REQUEST_CHANGES on an own PR; treating this as blocking comments. Do not mark ready until the fuzzy gate, the XSS oracle, and the #370 plan sync are fixed.


The branch is no longer documentation-only. After the second commit it adds host gates plus a production filter change: `fwlive-log-filter.sh` takes a dump directory and rpcd always passes `/tmp`. CI is green. CodeRabbit skipped the draft.

## Must fix

1. **Fuzzy PO gate exits 0** — `tests/fwlive-i18n.test.js` logs `[FAIL]` for fuzzy entries but omits them from `failures++`. `testFuzzyParser()` never drives that counter. A later `msgmerge` can ship unreviewed translations with a green suite.
2. **XSS oracle skips the tbody `innerHTML` sink** — production `renderAllRows` writes `host.innerHTML = ''`. The new test only walks children. The recording harness does not parse HTML, so `host.innerHTML = hostileMessage` would still pass.
3. **Plan file still contradicts #370** — this PR’s body already listed the L→R sync, R7 union rule, R1→#371, R6 merge, and R3/R9 in minimum-ready. `docs/developer/test-approach.md` was not updated. `security-review.md` then overclaims “fuzzy PO rejection” and “real reply-shape handling”.

## Should fix

4. **ipk gate uses GNU `find -printf`** — breaks the documented macOS `fwlive-test.sh` path. Use a portable `find | sed` listing; keep the module `sort -z` loop POSIX too.
5. **Title/body are stale** — summary still says documentation-only. Keep R9a’s source-shaped fixture as an explicit residual (`FWLIVE_REQUIRE_IPK` is unset in CI).

## Still open vs #370 (do not treat as done)

| Item | State |
| --- | --- |
| R1 nf_log | Correctly not here; lives in #371 |
| R3 `applyHash` | Still zero test references |
| R4b session ACL | Host JSON only |
| R6 view contract | Fixture fields without render assertions; no IPv6/hostile canned row |
| R7 union WAN lookup | Docs still treat it as undecided; no implementation |
| R8 / R9b | Manual / apk, as decided |
| R9a real ipk | Fixture fallback unless `FWLIVE_REQUIRE_IPK=1` |

What looks solid: filter dump-dir seam, ACL list vs read/write isolation (R4a, not R4b), UCI set/delete fail-closed, PATH-shadowed flock, rules byte cap, empty-ring `messages_received: 0`, IPv6 `name =` / `Address N` parse, and the 32-name resolve cap.

Fix 1–3 before Ready. Trigger CodeRabbit only after that push.

.github/workflows/fwlive-test.yml
```diff
@@ -36,6 +36,29 @@ jobs:
...
+        run: |
+          mkdir -p .ci-sdk-cache/dl .ci-sdk-cache/feeds/24.10.8
+          sudo chown -R 1000:1000 .ci-sdk-cache
+          sudo chmod -R u=rwX,g=rX,o=rX .ci-sdk-cache
+          OWRT_MAKE_JOBS=4 ./scripts/docker-sdk.sh build --target x86-64 --version 24.10
+      - name: Inspect real ipk payload
+        run: |
+          version="$(sed -n 's/^PKG_VERSION:=//p' openwrt-feed/luci-app-fwlive/Makefile)"
+          ipk="$(find "$GITHUB_WORKSPACE/out/x86_64/24.10.8/fwlive" \
+            -type f -name "luci-app-fwlive_${version}*.ipk" -print -quit)"
```
lucas-albers-lz4: Non-blocking: both paths hardcode `24.10.8` while the build is `--version 24.10`.

`sdk_matrix_version_label` already maps that pair. Source it and use the label for the feeds dir and the `out/` find, otherwise the next 24.10 patch bump fails this required job independently of `docker-sdk.sh`.

`find … -print -quit` is GNU-only. This runner is Ubuntu so it works, but `publish-packages.yml` already avoids `-quit` for that reason. `find … -print | head -n 1` is enough here.

docs/developer/test-approach.md
```diff
@@ -184,16 +184,86 @@ necessary history.
...
+   preserve unrelated state where rollback or configuration is involved.
+5. Treat installed-session and DOM claims as boundary tests. The ACL item
+   (R4a/R4b, formerly L1)
+   needs a real LuCI/uhttpd `/ubus` session login for a role with the read ACL
+   and a role without it; a root `ubus call` is not ACL evidence. The renderer
+   item (R2, formerly L4) should use the LuCI-accurate `E()` harness with hostile message
+   content, allow only intentional empty-container clearing, and assert that
+   the hostile value never appears in an `innerHTML` write while appearing as
+   text. The browser fixture item (R6, formerly L5) needs assertions for the changed
+   backend, adaptive/degradation metadata, IPv6 row, and hostile row—not only
```
lucas-albers-lz4: Non-blocking leftover: this still describes R4a and R4b as one session-login requirement. R4a is the host `list`/ACL parity test and is done. R4b is the installed LuCI session check and is still open.

R6’s “not only fields copied into a mock reply” requirement is still unmet; that is acceptable now that `security-review.md` only claims representative fixtures, not renderer proof.

lucas-albers-lz4 (commented):
Grok follow-up on `758906a` + `1c5b721`. Previous blocking findings are addressed. CI is green, including `test-ipk-payload` (2m53s after the uid-1000 cache prep).

## Closed from the last round

| Finding | Resolution |
| --- | --- |
| Fuzzy PO gate exits 0 | `if (total > 0) failures++` |
| XSS oracle skipped tbody | `collectInnerHTMLWrites(sinkBody)` must equal `['']` |
| GNU `find -printf` / `sort -z` | portable `find \| sed \| LC_ALL=C sort` |
| Plan still L1–L9 / contradicted #370 | R-numbers, R7 union rule, R1→#371, R6 merge, R3/R9 in minimum-ready |
| Security-review overclaim | now “representative reply-shape fixtures” |
| Stale PR title/body | updated |
| `test-ipk-payload` chown fail | `1c5b721` sudo chown/chmod to uid 1000 |

Bugbot found no new bugs on the fix commit.

## Remaining (non-blocking)

1. **`test-ipk-payload` hardcodes `24.10.8` twice** (feeds dir and `out/` find). `scripts/lib/sdk-matrix.sh` already maps `24.10` → `24.10.8`. A patch bump will fail this required job while `docker-sdk.sh --version 24.10` keeps working. Derive the label from `sdk_matrix_version_label`. Also drop GNU `find -print -quit`; `publish-packages.yml` already avoids that.
2. **Cache prep is thinner than the publish job.** Missing the `#208` symlink-path reject. Fine on a fresh runner today; copy that guard if this job stays in the per-PR path.
3. **Item 5 still lumps R4a with R4b.** Host ACL/method parity is R4a and is done. Session login remains R4b and is still open. R6 render assertions (IPv6/hostile/backend) remain open by choice — that is consistent with the walked-back ledger wording.
4. **`testFuzzyParser()` still does not drive `main()`’s exit status.** The live `failures++` fix is the real gate. A temp `.po` in the catalog loop would pin a future regression of that line.

R1 stays in #371. R3 `applyHash`, R4b, R7 implementation, and R8/R9b remain the #370 leftovers, not this PR’s remaining defects.

I would not block Ready on 1–4. Keep the PR draft until you want CodeRabbit; it is still skipped.

lucas-albers-lz4 (commented):
Grok follow-up on `34027df` + `5dee075`. The earlier review issues are resolved. The 24.10-IPK / 25.12-APK matrix is remaining work and does not reopen those findings.

## Closed

| Finding | Commit |
| --- | --- |
| Fuzzy gate exit 0 | `758906a` (`if (total > 0)`) |
| Fuzzy mutation did not drive the catalog gate | `5dee075` (`inspectCatalog` + `failure !== 1`) |
| XSS oracle skipped tbody | `758906a` |
| GNU `find -printf` / `sort -z` | `758906a` |
| Plan L1–L9 / R7 / R1 / R4a vs R4b | `758906a` + `5dee075` |
| Hardcoded `24.10.8` and `find -quit` | `5dee075` (`sdk_matrix_version_label`, `sed -n '1p'`) |
| Missing `#208` symlink reject | `34027df` / `5dee075` |
| uid-1000 cache prep | `1c5b721` |

CI is green, including `test-ipk-payload`. CodeRabbit completed with no inline findings.

## Remaining (explicitly not blocking this confirmation)

The job is still a 24.10 IPK gate. Selecting package format per release (IPK for 24.10, APK for 25.12) and validating that artifact type is new work, as written. R3 `applyHash`, R4b session ACL, R6 renderer assertions, R7 implementation, and R8 stay #370 leftovers.

lucas-albers-lz4:
Follow-up resolved in b6ce5fa:

- The SDK matrix now maps 24.10 and older supported lines to IPK, and 25.12/snapshot to APK.
- The existing test-ipk-payload job id is preserved for branch protection, but it now builds and inspects both 24.10 IPK and 25.12 APK artifacts.
- APK payloads are extracted with the matching digest-pinned SDK image, running as the runner UID, and the shared inspector checks the same shipped files, ACL, menu, and modes for both formats.
- Local full suite, real IPK, real APK, shellcheck, and actionlint all pass.

The completed CodeRabbit review reported no actionable comments. Fresh CI for b6ce5fa is pending.

lucas-albers-lz4 (commented):
Grok review of `b6ce5fa`. The IPK/APK matrix work is implemented and executed.

`sdk_matrix_package_format` maps 24.10 → ipk and 25.12 → apk. The stable `test-ipk-payload` job builds both cells, extracts the 25.12 apk with the pinned SDK host `apk`, and runs the format-aware inspector. CI passed both artifacts:

- `luci-app-fwlive_0.1.44-r1_all.ipk`
- `luci-app-fwlive-0.1.44-r1.apk`

Bugbot found no new bugs. I would not block on leftovers.

Non-blocking: the host-suite `command -v apk` fallback is the wrong tool class (Alpine apk vs OpenWrt apk-tools v3). CI is safe because it always passes `FWLIVE_PAYLOAD_DIR` from the SDK extractor. Fail closed without that dir rather than calling host `apk`. `test-approach.md` still lists R9b as unenforced package-format evidence; payload layout for both formats is now in this job. Remaining R9b is installed `apk add`/`apk del`, not archive inspection.

CodeRabbit is rate-limited on this push.

tests/fixtures/luci-view-boot.js
lucas-albers-lz4: Confirmed disposition: this PR now claims representative reply-shape fixtures only, not full renderer assertions. The backend, adaptive, IPv6, and hostile-row renderer assertions remain explicit #370 follow-up work. The implementation requested by this thread is therefore intentionally deferred, and this review thread can be resolved without treating R6 as complete.

lucas-albers-lz4:
Review-thread closure pass completed for head 7d1d604:

- All 8 inline review threads are resolved.
- Fuzzy-gate, XSS-oracle, documentation, portable package inspection, SDK-label, cache-symlink, and R4a/R4b wording findings are addressed by the pushed commits.
- R6 is explicitly resolved as a scope disposition: this PR claims representative fixtures only; full backend, adaptive, IPv6, and hostile-row renderer assertions remain follow-up work under #370.
- The latest CI run is green, including the 24.10 IPK and 25.12 APK payload checks.
- CodeRabbit status is rate-limited, not a code failure; its last completed review had no actionable comments.

Issue #375, opened 2026-09-19
ci(lint): shipped-JS ESLint gate enforces nothing — AMD-wrap processor lints a virtual filename no config matches


lucas-albers-lz4:
## Summary (facts)

`npm run lint:js` — run by the `test` job of `fwlive-test` — currently enforces **no ESLint rule** on any shipped LuCI module. All 15 modules under `htdocs/luci-static/resources/fwlive/` plus `view/status/fwlive.js` are processed with no configuration attached.

## Mechanism

`eslint.config.js` on master attaches a processor and a rule set to the shipped-JS file set:

- `74-79` — `luciAmdWrap.preprocess()` wraps each file in an IIFE and returns the virtual filename `name + '.wrapped'`
- `112` — `files: [...]` (globs end in `.js`)
- `116-123` — `plugins: { 'luci-amd-wrap': ... }`, `processor: 'luci-amd-wrap/wrap'`
- `124-153` — `languageOptions` (`sourceType: 'script'` + globals) and the full rule set

ESLint **flat config matches configuration per filename**. The processor's virtual filename (`chips.js.wrapped`) does not match the block's `files` glob, and no other block matches it, so the processed text is linted with no rules and no globals.

`--print-config` shows the asymmetry directly:

- `openwrt-feed/.../fwlive/chips.js` → `processor: luci-amd-wrap/wrap`, `no-undef: [2, {typeof: false}]`, 67 rules
- `openwrt-feed/.../fwlive/chips.js.wrapped` → no processor, no rules

## Reproduction — local (master)

1. Add a file under `htdocs/luci-static/resources/fwlive/` that calls an undefined global, then run `npm run lint:js` → exit 0, no output.
2. Temporarily add a config block matching `**/*.js.wrapped` carrying the same `languageOptions`/`rules` as the shipped-JS block, then run the same command against the unmodified shipped tree → 23 findings:
   - **10 × `no-undef`**: `getComputedStyle` ×3, `renderPolicy` ×2, `performance` ×2, `renderScheduler` ×1, `pollCoordinator` ×1, `Event` ×1. The three `render*`/`poll*` names are `'require … as …'` aliases declared at `view/status/fwlive.js:22-24`; the others are browser globals. All are missing from the config's globals/alias lists, not from the code.
   - **13 × `no-unused-vars`**: unused `catch (e)` bindings — `fwlive/logging.js` (2 sites), `view/status/fwlive.js` (11 sites).

[...]

PR #374, opened 2026-09-19
probe: CodeRabbit tool-layer probe (do not merge)

lucas-albers-lz4:
Scratch PR — **do not merge**. Purpose: determine what CodeRabbit's sandboxed tool layer actually executes on this repo.

Two deliberate violations:
1. `htdocs/.../fwlive/chips.js` — undefined global + missing semicolon. This file is covered by the `luci-amd-wrap/wrap` processor in `eslint.config.js`. Expected from the repo's own config: `no-undef` + `semi`. (Locally it reproduces: the repo's `npm run lint:js` reports **nothing** — the processor's virtual filename `*.js.wrapped` matches no config block, so no rules are applied to shipped JS. Adding a block for `**/*.js.wrapped` surfaces 11 `no-undef` errors.)
2. `root/usr/libexec/fwlive-adaptive-cap.sh` — unquoted expansion of an unassigned variable (shellcheck SC2086/SC2154).

Questions this answers:
- Does CodeRabbit's ESLint tool apply this repo's ESLint config (processor included), or does it silently skip/produce nothing?
- Does the shellcheck tool surface findings on this repo at all?
- Are third-party tools available on this plan (the tools docs carry an Essentials plan badge, yet Biome aborted on a previous PR while every workflow here was `action_required`)?

## Summary by CodeRabbit

* **Chores**
  * Added development validation checks to improve configuration and code-quality verification.
  * Expanded linting coverage for browser-based and wrapped JavaScript files.
  * Added support for additional browser globals and application view references during validation.

* **Refactor**
  * Clarified handling of intentionally unused error details without changing application behavior.
  * Preserved existing consent, storage, logging, and polling behavior.

PR #377, opened 2026-09-19
fix(ci): make the shipped-JS ESLint gate effective

lucas-albers-lz4:
## Problem

The shipped-JS ESLint gate enforced nothing. `eslint.config.js` attaches its rules to files matching `…/fwlive/**/*.js`, but the AMD-wrap processor hands ESLint a **virtual** filename (`<name>.js.wrapped`), and flat config matches configuration per filename — so no rule or global ever reached the processed code. `npm run lint:js` exited 0 on files containing undefined globals.

Evidence (full detail in the linked issue): a file calling an undefined global passes `npm run lint:js`; `--print-config` shows the processor and 67 rules for `chips.js` and nothing for `chips.js.wrapped`; and on the probe PR #374, `lint:js` passed a planted undefined global while CI failed only later on `prettier --check` and on an unrelated shell probe.

## Change

- Hoist the shipped-JS file set, `languageOptions` and rule set into named constants (`shippedJsFiles`, `shippedJsLanguageOptions`, `shippedJsRules`) so the rules have one definition.
- Attach those same options and rules to the processor's virtual filename via a second block (`files: ['**/*.js.wrapped']`), which is the standard pattern for processors that lint generated text (e.g. eslint-plugin-markdown).
- Complete the two config lists that the now-live rules immediately exposed: `getComputedStyle`, `performance`, `Event` added to the browser globals; `renderPolicy`, `pollCoordinator`, `renderScheduler` added to the LuCI `'require … as …'` aliases (declared at `view/status/fwlive.js:22-24`).
- Rename 13 unused `catch (e)` bindings to `catch (_e)` (`fwlive/logging.js` ×2, `view/status/fwlive.js` ×11). The rule's own `caughtErrorsIgnorePattern: '^_'` exists for exactly this; the bindings are unused, so behaviour is unchanged.

## Verification

- **Negative control:** with the fix in place, injecting an undefined global + missing semicolon into a shipped module makes `npm run lint:js` report `no-undef` + `semi` (2 errors) — before the fix the same file passed.
- **Full lint stack green:** `npm run lint` (eslint + prettier `--check` + stylelint) passes with all 23 newly surfaced findings resolved (10 missing-global + 13 unused-binding).
- **Tests:** `fwlive-modules-smoke`, `fwlive-view-logging-toggle`, `fwlive-view-poll-guard`, `fwlive-view-fetch-budget`, `fwlive-theme-tint`, `fwlive-i18n` all pass.

## Notes

- Behaviour-preserving for the shipped modules: the only source edits are unused catch-binding renames.
Refs #375

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when browser storage is unavailable, such as in private browsing mode.
  * ESLint checks now correctly apply to wrapped JavaScript files, including undefined-variable detection.

* **Tests**
  * Added regression coverage to verify linting of wrapped JavaScript files.
  * Updated the test runner to include the new validation.


eslint.config.js
```diff
@@ -121,33 +164,15 @@ module.exports = [
...
-			'no-var': 'off',
-		},
+		languageOptions: shippedJsLanguageOptions,
+		rules: shippedJsRules,
+	},
+	{
+		/* The AMD-wrap processor lints a virtual `<name>.js.wrapped` file, not the
+		 * module on disk. Attach the same environment and rules to that virtual
+		 * filename, otherwise no rule (no-undef, semi, no-eval, …) ever applies. */
+		files: ['**/*.js.wrapped'],
```
lucas-albers-lz4, line 196: This is the actual gate. After the unused-`catch` cleanups, dropping this block would make `lint:js` green again with no remaining in-tree findings. A `--print-config` assertion on a `*.js.wrapped` path (that `no-undef` is `error`) pins the #375 failure without planting a shipped-tree violation.

lucas-albers-lz4 (commented):
## Review

The AMD-wrap / flat-config diagnosis is correct, and attaching the shared `shippedJsLanguageOptions` / `shippedJsRules` to `**/*.js.wrapped` is the right processor pattern. The extra globals match real use (`getComputedStyle`, `performance`, `Event`, plus the `require … as …` aliases). The 13 `catch (_e)` renames are unused bindings only; `poll-coordinator.js` still uses `catch (e)` for `Promise.reject(e)` and is correctly left alone. Independent of #379/#380.

**Non-blocking:** after those unused-binding cleanups, deleting the `.wrapped` block would make `lint:js` exit 0 again — the original #375 failure mode. I'll push a `--print-config` check that the virtual filename has `no-undef: error` so this cannot go silent a second time.

lucas-albers-lz4:
Pushed `375e281` on this branch: `tests/fwlive-eslint-amd-wrap.test.sh` asserts `eslint --print-config` on `chips.js.wrapped` still has `no-undef` at error, wired from `scripts/fwlive-test.sh` next to `lint:js`.

lucas-albers-lz4:
## Independent review — verdict: APPROVE (head verified: `a3d36c0`, incl. the follow-up rule fold)

Diagnosis is correct (flat config matches per filename; the AMD-wrap processor's virtual `*.js.wrapped` files matched no rule block, so the gate enforced nothing) and the fix is the standard two-block pattern.

**Empirically verified at `375e281` (pre-fold) and `a3d36c0` (post-fold):**

- **Negative control:** appending `undefinedGlobalProbe();` to `chips.js` makes `lint:js` fail with `no-undef` + `no-unreachable` (exit 1); restored → green. The gate genuinely enforces now.
- **Regression pin:** `tests/fwlive-eslint-amd-wrap.test.sh` passes — `--print-config` on `chips.js.wrapped` shows `no-undef = error` (67 rules total). shellcheck clean.
- **New browser globals are all real usages:** `getComputedStyle` (fwlive.js:513+), `performance` (:1045), `new Event` (:1707). New aliases are declared at fwlive.js:22-24 (`'require fwlive.poll-coordinator as pollCoordinator'` etc.).
- **13 `catch (e)` → `catch (_e)` renames** match `caughtErrorsIgnorePattern: '^_'` exactly (2 in logging.js, 11 in fwlive.js). Behaviour-preserving.

**Follow-up commit `a3d36c0`** (folded from the #380 review): `no-new-func` + `no-implied-eval` added to `shippedJsRules`. Probe verification:

| shape | result |
|---|---|
| `Function('return 1')` | `no-new-func` ✅ |
| `new Function('return 1')` | `no-new-func` ✅ |
| `setTimeout('doThing()', 100)` | `no-implied-eval` ✅ |
| `window.eval(…)` | `no-eval` (already) ✅ |
| shipped tree | clean — `npm run lint:js`, regression test, and full `npm run lint` all green |

These two rules close ESLint's half of the dynamic-code gate; the residual shapes (`globalThis.Function`, `window['Function']`) are out of `no-new-func`'s scope by design and are noted on #380.

eslint.config.js
lucas-albers-lz4: Pinned in `375e281` (`no-undef` on `chips.js.wrapped`). Follow-up `52a6395` also asserts `no-new-func` / `no-implied-eval` / `no-restricted-properties` on that virtual path, plus a wrap-path probe that `globalThis.Function` / `window.Function` / `window["Function"]` fire.

lucas-albers-lz4:
Addressed CodeRabbit’s nit on `a3d36c0`: `no-new-func` does not see member access, so `52a6395` adds `no-restricted-properties` for `globalThis.Function` and `window.Function` (direct and `window['Function']`). `tests/fwlive-eslint-amd-wrap.test.sh` now asserts those rules on the wrap path and plants a 3-hit probe.

The Biome “Illegal return statement” abort is the LuCI AMD `return` that #379 turns off (`biome.enabled: false`). Not a shipped syntax error.

PR #379, opened 2026-09-19
ci(coderabbit): correct tool keys, drop nonexistent bullseye, disable biome

lucas-albers-lz4:
## What changes

`.coderabbit.yaml` (under `reviews.tools`):

- `ast_grep: true` → `ast-grep: true`. The documented key is hyphenated; the underscore spelling is not a schema key, so the entry configured nothing. CodeRabbit's own effective configuration (`@coderabbitai configuration`) confirms it: `ast-grep` appears with `# Source: defaults`, `ast_grep` appears nowhere. Fixing the key is a prerequisite for ever passing `rule_dirs`/`util_dirs`/`packages` for this tool.
- `bullseye: true` → removed. No such tool exists in CodeRabbit's configuration reference (50 documented tools), the docs index, the tool catalog, or this repo's effective configuration (zero occurrences). The intent recorded in the comment — a security add-on for the rpcd/frontend boundary — is served by semgrep/opengrep (both enabled by default) plus the existing `path_instructions` for `openwrt-feed/.../libexec/**`.
- `biome: enabled: false` added. Biome 2.5.11 cannot parse a LuCI AMD module: every shipped file ends in a top-level `return baseclass.extend({ … })`, which LuCI supports by evaluating the module inside a function wrapper. Biome aborts on each of the 15 shipped modules (`Illegal return statement outside of a function`) and has never produced a finding in this repo. Its only observable output was the `Some tools did not complete` banner on LUCI-touching PRs. Oxlint and ESLint parse the same files; CI's `lint:js` is the gate.

`docs/developer/coderabbit.md`: a new *Tool-level configuration* subsection recording all three decisions, plus the fact that ast-grep's essentials package contributes nothing here (553 rules, 7 JavaScript rules all aimed at Node/Express libraries, zero shell rules) so any value must come from repo-authored rules — tracked as an investigation, not implied by this config.

Net effect: one visible change (the Biome abort banner stops appearing) and two removals that were already no-ops. No change to `profile`, `request_changes_workflow`, `auto_review`, `pre_merge_checks`, or any `path_instructions`.

## Verification

- `.coderabbit.yaml` parses and resolves to `tools = {'ast-grep': True, 'biome': {'enabled': False}}`.
- Config-only + docs change; no shipped code touched, no CI gate affected.
- Post-merge check: `@coderabbitai configuration` should show `biome: enabled: false` with `# Source: Repository YAML`, `ast-grep` sourced from the repo YAML, and no `bullseye`.

[...]

Issue #376, opened 2026-09-19
ci(coderabbit): tool-layer disposition for fwlive — configure now, discards, and open investigations


lucas-albers-lz4:
## Context

Audit of CodeRabbit's tool layer on fwlive, done as a pilot before touching any other repo. Findings that frame every decision below:

- Tools run, but produce nothing useful here. An induced-violation PR (#374) planted an undefined global + missing semicolon in a shipped LuCI module and an unquoted unassigned variable in a libexec shell script. **No tool surfaced either**; both were caught by the LLM review. The only tool-layer output in the repo's history is Biome's abort banner (`Illegal return statement outside of a function`) on LuCI modules.
- Volume is already low: 0–2 CodeRabbit inline comments per PR across the last 12 PRs. Noise is not a problem to solve.
- Plan: the account is the free tier; CodeRabbit grants public repositories free access and reports `Plan: Advanced` for this repo. So paid-tier gates are satisfied here via the OSS grant — verify per-feature rather than assuming.

## Configure now (decided)

- `reviews.tools.biome.enabled: false`. Biome 2.5.11 cannot parse a module with a top-level `return` in any mode (`.js`/`.cjs`/`.mjs`/`type: module` all fail), it has no parser option for it, and it is 0-for-15 on the shipped modules. Oxlint (default-on) and ESLint both parse the same files.
- `ast_grep: true` → `ast-grep: true`. The documented key is hyphenated; `ast_grep` is not a schema key, so the entry at `.coderabbit.yaml:28` is inert (and `rule_dirs` could not be passed through it).
- Delete `bullseye: true` (`.coderabbit.yaml:29`). `bullseye` does not appear anywhere in CodeRabbit's configuration reference (50 documented tools).

## Discarded (with reasons)

- Converting shipped LuCI modules to ESM for Biome — the loader `eval`s the source inside a function wrapper and consumes its return value; `export` is a SyntaxError there.
- Adding a `biome.json` purely to suppress the tool — CodeRabbit's Biome integration accepts no repo config (`enabled` only), and the repo does not use Biome.
- Listing guideline filenames in `reviews.path_instructions` — documented antipattern: that reviews them as code instead of applying them as criteria.
- `reviews.path_filters` excluding `htdocs/**` — would remove LLM review from the highest-risk surface.
- `reviews.request_changes_workflow: true` / pre-merge checks at `error` — blocks merges on bot opinion; the repo's stated stance is that CI is the gate.
[...]

Issue #138, opened 2026-08-10
tests: E() stub never renders, so DOM-injection regressions are invisible to CI


lucas-albers-lz4:
The module loader stubs `E()` as a plain object constructor:

```js
function fakeE(tag, attrs, children) {
	return { tag: tag, attrs: attrs || {}, children: children };
}
```

Nothing is ever rendered, so every renderer test asserts on a descriptive object rather than on resulting DOM. Any bug where a value reaches an HTML sink instead of a text node is structurally invisible to the suite — the stub does not model the upstream LuCI behaviour that distinguishes the two.

`tests/fwlive-modules-smoke.test.js` does check that hosts are cleared (`assert.strictEqual(body.innerHTML, '')`), but never what gets written afterwards.

## Proposal

Add a LuCI-accurate `E()` test harness alongside `fakeE`:

- Minimal `Node` class with an `innerHTML` **setter that records every write**, plus `appendChild` / `setAttribute`.
- `append()` and `create()` ported verbatim from upstream `luci.js` (`modules/luci-base/htdocs/luci-static/resources/luci.js`, `dom.append` / `dom.create`) so array-vs-bare-string child handling matches production exactly.
- A test that renders rows and filter chips with markup-bearing field values and asserts **no recorded `innerHTML` write contains `<`**.

Keep `fakeE` for the existing structural assertions; the new harness is for encoding safety specifically.

This is the regression gate for #137. A working harness of this shape already exists from the audit and can be adapted directly.

## Value beyond the current fix

The array-vs-string distinction is easy to reintroduce in any future renderer edit, and it is invisible in review. A test that renders is the only durable guard.

lucas-albers-lz4:
A working harness for this now exists and has been validated, but it lives in the **private advisory** and its temporary private fork rather than here — it renders markup-bearing values, so posting it publicly would disclose the related advisory ahead of its fix.

## What exists

Two drop-in files on branch `fix/137-output-encoding` in the advisory's private fork:

| File | Role |
|---|---|
| `tests/lib/luci-dom.js` | `Node` class whose `innerHTML` **setter records every write**, plus `append`/`create` ported verbatim from upstream LuCI so array-vs-bare-string child handling matches production exactly |
| `tests/fwlive-dom-encoding.test.js` | Renders the real `log`/`links`/`table`/`chips`/`logging` modules through that `E()` and asserts no recorded write contains `<` |

[...]

PR #380, opened 2026-09-19
ci(invariants): ast-grep rules for shipped JS (CI gate + CodeRabbit rule_dirs)

lucas-albers-lz4:
## What this adds

Three ast-grep rules that encode the repo's own documented invariants for the shipped LuCI surface, wired so the **same rule directory** runs in CI and is read by CodeRabbit.

**Rules** (`scripts/ast-grep-rules/`, config in `sgconfig.yml`):

| rule | enforces |
|---|---|
| `fwlive-e-children-must-be-array` | `security-model.md` §1 — a provably string child (string literal, template literal, concatenation, or a `_()` call) in `E(...)` is an `innerHTML` sink upstream; use the array form |
| `fwlive-html-sink-nonliteral` | no non-literal assignment into `innerHTML`/`outerHTML`/`insertAdjacentHTML`, no `document.write` (literal clears such as `host.innerHTML = ''` stay allowed) |
| `fwlive-no-dynamic-code` | no `eval` / `new Function` / `Function` in shipped modules |

**Wiring**

- `scripts/fwlive-ast-grep.sh` — gate wrapper, pinned to **ast-grep 0.45.3** (the version CodeRabbit runs), called from `scripts/fwlive-test.sh` next to the shellcheck gate.
- `.github/workflows/fwlive-test.yml` — `pipx install 'ast-grep-cli==0.45.3'` alongside ruff.
- `tests/fwlive-ast-grep-rules.test.sh` — self-test: rules must be silent on the shipped tree, must fire on a planted violation, and must not flag compliant shapes. Guards rule rot.
- `.coderabbit.yaml` — `reviews.tools.ast-grep`: `rule_dirs: ['scripts/ast-grep-rules']` and `essential_rules: false` (the essentials package is 553 rules aimed at Java/Python/C++ and Node/Express libraries, with **zero** shell rules, so it contributes nothing here).
- Docs: `security-model.md` §1 records the enforcement; `build-and-test.md` lists the gate.

## Zero-noise bar (why this shouldn't page anyone)

Authoring started from two false-positive classes that a naive rule set produces, both measured and fixed:

- `E('thead', {}, E('tr', {}, []))` — a **Node** child is safe upstream (`this.elem()` → `appendChild`); the rule excludes call children.
- `E('span', { … }, parts)` — an **identifier** child is unknowable statically (array or Node); the rule only fires on provably string children.
- `host.innerHTML = ''` — the clear idiom; literal RHS is excluded.

Verification:

- `ast-grep scan` on the current shipped tree: **0 findings**.
- Planted violations: all three rules fire.
- Compliant shapes (array form, Node child, identifier child, literal clear): silent.
- `shellcheck -s bash` clean on both new scripts; `npm run lint` unaffected.

## Scope and non-goals

- Shipped LuCI JS only (rule `files:` globs). `tests/fixtures/**` is deliberately out of scope: the harness emulates LuCI's own `innerHTML`/`new Function` loader internals.
- **Bashisms stay with shellcheck.** ast-grep's Bash grammar rejects those patterns outright (`[[ … ]]`, `source $f`, `declare`, arrays), and the shellcheck discovery gate already covers shipped libexec/rpcd.
- No taint analysis: an untrusted identifier reaching a sink is not statically decidable by AST patterns — that class remains a review/semgrep concern, which is why `essential_rules` is off rather than replaced by a noisy rule set.

## Stacking

Stacked on #379 (config hygiene) so the `.coderabbit.yaml` hunk applies cleanly; the diff below is relative to that branch.

Refs #376.

## Summary by CodeRabbit

* **Security**
  * Added automated checks for unsafe HTML injection, dynamic code execution, and unsafe `E()` child rendering patterns in shipped JavaScript.
  * Added regression coverage to confirm compliant code passes without false positives.

* **Testing**
  * Extended the test workflow to run the new static security checks.
  * Standardized validation on ast-grep version 0.45.3.

* **Documentation**
  * Documented the new static security checks and their role in the parser test suite.


lucas-albers-lz4:
## Multi-Review Synthesis (Claude Sonnet, GLM-5.2, Grok)

A parallel review was conducted across three subagent models (**Claude Sonnet**, **GLM-5.2**, and **Grok**), verified by running pinned `ast-grep 0.45.3` locally and executing probe tests against real and planted AST shapes. The reviews were in unanimous agreement.

---

### 1. Critical CI Issue: Rule-rot self-test is never executed

`tests/fwlive-ast-grep-rules.test.sh` was added specifically to guard against rule rot (asserting that rules fire on planted violations, pass on clean code, and avoid false positives).

- **Current state**: `tests/fwlive-ast-grep-rules.test.sh` is **not referenced anywhere** in `scripts/fwlive-test.sh`, `.github/workflows/fwlive-test.yml`, or `package.json`.
- **Impact**: CI only runs `scripts/fwlive-ast-grep.sh` against the clean tree. If a rule breaks, becomes a no-op, or has a broken glob, CI exits 0. The rule-rot guarantee advertised in the test header is unfulfilled.
- **Fix**: Wire the test script into `scripts/fwlive-test.sh`:
  ```bash
  echo "== fwlive invariant rules, shipped JS (ast-grep) ==" >&2
  bash "$ROOT/scripts/fwlive-ast-grep.sh"
  bash "$ROOT/tests/fwlive-ast-grep-rules.test.sh"
  ```

---

### 2. AST-Grep Rule Blind Spots & Fixes

#### A. `fwlive-e-children-array.yml`
- **Identifier attributes bypass**: `constraints: ATTRS: kind: object` restricts matches strictly to object literals. Passing attrs as an identifier/variable (such as `E('td', timeAttrs, ...)` already used in `table.js:140`) completely bypasses the rule, even with an unsafe child.
- **Two-argument calls omitted**: In LuCI `dom.create` (`tests/lib/luci-e-harness.js`), passing a non-object 2nd argument treats it as children (`data = attr, attr = null`). Therefore, `E('div', 'literal')` and `E('span', _('text'))` are `innerHTML` sinks, but `pattern: E($TAG, $ATTRS, $CHILD)` requires 3 arguments and misses them.
- **Recommended fix**: Drop `ATTRS: kind: object` and support both arities:
  ```yaml
  rule:
    any:
      - pattern: E($TAG, $CHILD)
      - pattern: E($TAG, $ATTRS, $CHILD)
  constraints:
    CHILD:
      any:
        - kind: string
        - kind: template_string
        - kind: binary_expression
        - pattern: _($$$)
  ```
  And add test cases for 2-arg calls and identifier attrs in `tests/fwlive-ast-grep-rules.test.sh`.

#### B. `fwlive-html-sink.yml`
- **Missing `+=` augmented assignment**: `$EL.innerHTML += $VAL` and `$EL.outerHTML += $VAL` produce AST `augmented_assignment_expression` and are not caught by the `=` pattern.
- **Recommended fix**: Add `$EL.innerHTML += $VAL` and `$EL.outerHTML += $VAL` to the `rule.any` list.

#### C. `fwlive-dynamic-code.yml`
- Catches direct `eval($$$)`, `new Function($$$)`, and `Function($$$)`.
- Minor: Consider adding `pattern: window.eval($$$)` / `pattern: globalThis.eval($$$)` to close common direct-window escapes.

---

### 3. Documentation Calibration (`docs/developer/security-model.md`)

The updated text states that invariant §1 is *"Enforced by `scripts/ast-grep-rules/fwlive-e-children-array.yml`"*. Because AST pattern matching cannot trace identifiers or function call returns to distinguish text nodes from strings, this rule guards a static provable subset. The runtime defense-in-depth and dynamic cases remain covered by the `#138` rendering regression harness. Consider phrasing this as *"Statically gated for provable string shapes by ... while dynamic/identifier flows remain guarded by the rendering regression harness."*

---

### 4. Stacking & Merge Sequence

1. Fix `ast-grep: enabled: true` in PR #379 so it is schema-valid on its own.
2. Merge PR #379 into `master`.
3. Rebase PR #380 onto updated `master` (which will cleanly supply `rule_dirs: ['scripts/ast-grep-rules']` and `essential_rules: false`).
4. Merge PR #380.

scripts/fwlive-test.sh
```diff
@@ -74,6 +74,9 @@ echo "OK: SPDX headers present on shipped JS/shell/Makefile" >&2
 echo "== fwlive shellcheck (libexec/rpcd) ==" >&2
 bash "$ROOT/scripts/fwlive-shellcheck.sh"
 
+echo "== fwlive invariant rules, shipped JS (ast-grep) ==" >&2
+bash "$ROOT/scripts/fwlive-ast-grep.sh"
```
lucas-albers-lz4, line 80: This only scans the clean tree. `tests/fwlive-ast-grep-rules.test.sh` (planted violations + compliant-shape silence) is not referenced from this script, the workflow, or `package.json`. A rule that becomes a no-op still exits 0. Call the self-test here:
```bash
bash "$ROOT/scripts/fwlive-ast-grep.sh"
bash "$ROOT/tests/fwlive-ast-grep-rules.test.sh"
```

scripts/ast-grep-rules/fwlive-e-children-array.yml
```diff
@@ -0,0 +1,22 @@
...
+  anything else into innerHTML - so E('td', attrs, value) is an HTML sink while
+  E('td', attrs, [ value ]) is not.
+  See docs/developer/security-model.md 1.
+files:
+  - openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/**/*.js
+rule:
+  pattern: E($TAG, $ATTRS, $CHILD)
+constraints:
+  ATTRS:
+    kind: object
```
lucas-albers-lz4: `kind: object` means identifier attrs bypass the rule. `table.js` already does `E('td', timeAttrs, [ … ])` — safe today because the child is an array; `E('td', timeAttrs, untrusted)` would not match.

LuCI `dom.create` also treats a non-object 2nd argument as children (`tests/lib/luci-e-harness.js`), so `E('div', 'literal')` and `E('span', _('x'))` are innerHTML sinks that this 3-arg pattern never sees.

Drop the ATTRS constraint and match both arities:
```yaml
rule:
  any:
    - pattern: E($TAG, $CHILD)
    - pattern: E($TAG, $ATTRS, $CHILD)
constraints:
  CHILD:
    any:
      - kind: string
      - kind: template_string
      - kind: binary_expression
      - pattern: _($$$)
```

scripts/ast-grep-rules/fwlive-html-sink.yml
```diff
@@ -0,0 +1,20 @@
...
+message: |
+  Untrusted values must reach the DOM as text nodes, never an HTML sink.
+  Assigning a non-literal to innerHTML/outerHTML/insertAdjacentHTML (or calling
+  document.write) bypasses that rule.
+  See docs/developer/security-model.md §1.
+files:
+  - openwrt-feed/luci-app-fwlive/htdocs/luci-static/resources/**/*.js
+rule:
+  any:
+    - pattern: $EL.innerHTML = $VAL
```
lucas-albers-lz4, line 13: `+=` is `augmented_assignment_expression` and does not match `= $VAL`. Add `$EL.innerHTML += $VAL` and `$EL.outerHTML += $VAL` (unused in the shipped tree today, but a one-line bypass of the sink rule).

docs/developer/security-model.md
```diff
@@ -66,6 +66,11 @@ sinks. The rendering regression harness
 `tests/` coverage) keeps it that way. The compliance sweep landed with
 [#137](https://github.com/lucas-albers-lz4/fwlive/issues/137).
 
+Enforced by `scripts/ast-grep-rules/fwlive-e-children-array.yml`: any provably
```
lucas-albers-lz4: AST patterns cannot follow identifiers or call returns, so this is not a full enforcement of §1. Prefer: statically gated for provable string shapes by the ast-grep rule; dynamic/identifier children remain guarded by the #138 rendering harness.

lucas-albers-lz4 (commented):
## Review

Stacking is correct (one commit on `ci/coderabbit-tool-config`). After #379 merges, rebase so the YAML hunk adds `rule_dirs` on top of the object form. This PR's `ast-grep: { enabled, rule_dirs, essential_rules: false }` is already schema-valid. Pinning ast-grep **0.45.3** matches CodeRabbit's current tool version.

**Blocking:** `tests/fwlive-ast-grep-rules.test.sh` is never invoked. `fwlive-test.sh` only runs `scripts/fwlive-ast-grep.sh` against the clean tree, so a rule that stops matching still exits 0.

**Rule holes** (not exploited in the current tree, but they undercut the “enforced” claim):

- `ATTRS: kind: object` skips identifier attrs (`E('td', timeAttrs, …)` already exists in `table.js`).
- Two-arg `E('div', 'literal')` / `E('span', _('x'))` is an innerHTML sink in LuCI `dom.create`; the pattern requires three arguments.
- `$EL.innerHTML = $VAL` misses `+=`.

`security-model.md` should say the rule gates **provably string** children; identifier/call/ternary flows stay with the #138 harness.

I'll push: wire the self-test, drop the ATTRS object constraint, add the 2-arg pattern and `+=`, plant those shapes, and calibrate the docs. Rebase onto the #379 schema fix.

lucas-albers-lz4:
Rebased onto #379 (`a47e65a`) and pushed `f423530`:

- `tests/fwlive-ast-grep-rules.test.sh` is now invoked from `scripts/fwlive-test.sh`
- E() rule matches 2-arg calls and identifier attrs; HTML-sink rule matches `+=`
- planted-violation test requires those extra hits; docs no longer claim full enforcement of §1

scripts/fwlive-test.sh
lucas-albers-lz4: Addressed in f423530: `tests/fwlive-ast-grep-rules.test.sh` is now invoked from `scripts/fwlive-test.sh` next to the gate.

scripts/ast-grep-rules/fwlive-e-children-array.yml
lucas-albers-lz4: Addressed in f423530: dropped the ATTRS object constraint and match both 2-arg and 3-arg E(); planted tests cover identifier attrs and 2-arg string children.

scripts/ast-grep-rules/fwlive-html-sink.yml
lucas-albers-lz4: Addressed in f423530: `$EL.innerHTML += $VAL` and `$EL.outerHTML += $VAL` are in the rule; the self-test requires ≥2 html-sink hits.

docs/developer/security-model.md
lucas-albers-lz4: Addressed in f423530: security-model.md now says the rule statically gates provable string shapes; identifier/call children stay with the rendering harness.

lucas-albers-lz4:
@coderabbitai review

lucas-albers-lz4:
## Independent review — verdict: APPROVE (head verified: `f423530`, pinned ast-grep 0.45.3)

**Core gates re-run locally:** `scripts/fwlive-ast-grep.sh` clean on the shipped tree (0 findings); `tests/fwlive-ast-grep-rules.test.sh` passes (all 3 rules fire on planted violations, ≥6 E() hits, ≥2 sink hits, silent on compliant shapes); shellcheck clean on both new scripts; version pin hard-fails on drift. The self-test being wired from `fwlive-test.sh` (f423530) closes the rule-rot gap properly.

**Independent bypass probe** (shapes beyond the self-test, scanned at 0.45.3 with these exact rules):

| shape | ast-grep result |
|---|---|
| `window.eval('…')` | silent ❌ |
| `(0, eval)('…')` | silent ❌ |
| `globalThis.Function('return 1')` | silent ❌ |
| `setTimeout('doThing()', 100)` | silent ❌ |
| `document.writeln(untrusted)` | silent ❌ |
| `E('td', {}, n > 1 ? 'yes' : 'no')` | silent ❌ |
| `E('td', {}, [1,2].join(','))` | silent ❌ |

### 🟡 Minor — `fwlive-no-dynamic-code` bypassable via member/indirect forms

`pattern: eval($$$)` matches only the bare identifier. Defense-in-depth now exists after #377's `a3d36c0` fold (`no-new-func` + `no-implied-eval`): ESLint catches `window.eval`, bare/`new Function`, and string-arg `setTimeout`. Residual for both gates: `globalThis.Function(...)`, `(0, eval)(...)`, `document.writeln`. Cheap hardening if wanted: add `- pattern: $OBJ.eval($$$)` to the rule and treat `writeln` like `write`; the identifier-taint cases (`(0, eval)`) are honestly out of scope per the PR's own non-goals.

### 🟢 Nits (fine to leave to the #138 harness)

- Provably-string ternary child `E('td', {}, n > 1 ? 'yes' : 'no')` doesn't fire (`conditional_expression` not in constraints) — slightly at odds with the "provably string shapes" wording in security-model.md §1.
- Bracket-notation sink `el['innerHTML'] = v` doesn't fire.
- Neither shape exists in the shipped tree today.

### ℹ️ Merge order

#377 and this PR both touch `scripts/fwlive-test.sh` (disjoint hunks — auto-merge expected, but #377 will need a rebase-check after #379 merges since this branch sits on #379's head, not #377's). Suggested sequence: #377 → #379 → #380.

**Bottom line:** rules do exactly what the PR body claims within their stated scope; the scope statements (no taint analysis, tests/fixtures out, bashisms to shellcheck) are honest and accurate. Approve.

lucas-albers-lz4:
@coderabbitai review

lucas-albers-lz4:
@coderabbitai rate limit

lucas-albers-lz4:
@coderabbitai review

scripts/ast-grep-rules/fwlive-html-sink.yml
```diff
@@ -0,0 +1,22 @@
...
+    - pattern: $EL.outerHTML = $VAL
+    - pattern: $EL.innerHTML += $VAL
+    - pattern: $EL.outerHTML += $VAL
+    - pattern: $EL.insertAdjacentHTML($POS, $VAL)
+    - pattern: document.write($VAL)
+constraints:
+  VAL:
+    not:
+      kind: string
```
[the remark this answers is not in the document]
lucas-albers-lz4: Addressed in `bc38885`: `document.write($$$)` / `document.writeln($$$)` are unconstrained (string literals included). The `VAL: not kind: string` exception stays only on innerHTML/outerHTML/insertAdjacentHTML so `el.innerHTML = ''` remains allowed. Self-test now requires ≥4 sink hits, including planted `document.write('literal HTML')` and `document.writeln('literal')`.

lucas-albers-lz4:
@coderabbitai review

PR #382, opened 2026-09-19
fix: make nf_log readiness family-aware

### LINKED ISSUE #371: defect: read_nf_log_backend treats uppercase NONE as present logger (fail-open)

lucas-albers-lz4:
Closes #371

## Summary
- Treat missing, empty, and case-insensitive NONE nf_log selectors as unavailable.
- Probe IPv6 family availability independently through /proc/net/if_inet6; IPv4 remains required by the supported WAN path.
- Drive status fields, blockers, readiness, and the enable gate from one computed state.
- Add backend/family fixture coverage and update operator/security documentation.

## Validation
- bash tests/fwlive-logging.test.sh
- ./scripts/fwlive-shellcheck.sh
- ./scripts/fwlive-test.sh

All pass.


tests/fwlive-logging.test.sh
```diff
@@ -42,6 +42,124 @@ case "$out" in
...
+	*'"nf_log_ipv4":true'*'"nf_log_ipv6":true'*'"ready":true'*) ;;
+	*) die "IPv4-ready/IPv6-absent status must be ready: $out" ;;
+esac
+case "$out" in
+	*'"nf_log_ipv6_missing"'*) die "absent IPv6 must not block status: $out" ;;
+esac
+out=$(enable_wan_logging)
+case "$out" in
+	*'"error":"nf_log_missing"'*) die "absent IPv6 must not block enable: $out" ;;
+esac
```
lucas-albers-lz4: **CONFIRMED — this does not prove IPv4-only enable succeeds.**

Unlike every other `enable_wan_logging` case in this file, this block does not stub `acquire_wan_log_lock` / `release_wan_log_lock` (or set `FWLIVE_WAN_LOG_LOCK_FILE` under the temp dir). The assertion only rejects `error:nf_log_missing`.

That is true for both of:
- `lock_failed` (non-root host, cannot mkdir `/etc/fwlive`)
- `ok:true,"changed":false` (root CI: lock is created at the production path, UCI stub already has `log=1`, so enable never stages)

Please stub the lock helpers like the later enable tests, keep the lock under `NF_LOG_WORK`, and assert `ok:true` for the IPv4-only case. The present-IPv6 `NONE` case below can stay on `nf_log_missing`, but it should also not open the production lock path.

tests/fwlive-logging.test.sh
```diff
@@ -42,6 +42,124 @@ case "$out" in
...
+out=$(build_logging_status_json)
+case "$out" in
+	*'"nf_log_ipv6":false'*'"ready":false'*'"nf_log_ipv6_missing"'*) ;;
+	*) die "present IPv6 with NONE must block status: $out" ;;
+esac
+out=$(enable_wan_logging)
+case "$out" in
+	*'"error":"nf_log_missing"'*) ;;
+	*) die "present IPv6 with NONE must block enable: $out" ;;
+esac
```
lucas-albers-lz4, line 205: **CONFIRMED — family matrix is missing the dual-ready cell, and the backend loop never drives status/enable.**

#371 asked the backend matrix to assert helper **and** status/blockers/ready, and the family matrix to include “both families present and both loggers ready ⇒ ready” (and a present IPv4 family with `NONE`/empty/missing blocking enable).

As written:
- the loop above only calls `read_nf_log_backend`
- this block covers IPv6-absent (effective-ready) and IPv6-present+`NONE` (blocked)
- there is no both-ready status assertion, and no IPv4-`NONE` status/enable assertion

Add those two cells against the same fixture seam so a later stub of `check_nf_log_ipv4` / `compute_nf_log_state` cannot desync the JSON fields from the enable gate.

openwrt-feed/luci-app-fwlive/root/usr/libexec/fwlive-logging.sh
```diff
@@ -425,17 +433,50 @@ read_nf_log_backend() {
...
+	family="$1"
+	case "$family" in
+		ipv4)
+			# Supported fwlive deployments require IPv4 for the WAN path.
+			return 0
+			;;
+		ipv6)
+			path="${FWLIVE_IPV6_AVAILABLE_PATH:-$NF_LOG_IPV6_AVAILABLE_PATH}"
+			[ -r "$path" ] || return 1
+			grep -q '[^[:space:]]' "$path" 2>/dev/null
```
lucas-albers-lz4, line 467: **Open — document the `if_inet6` signal as “IPv6 stack present”, not “IPv4-only WAN”.**

A content probe is the right seam, and missing/empty files correctly fail the family check. A populated file that only contains loopback `::1` still returns success here, so `check_nf_log_ipv6` will require a real `nf_log/10` backend.

That matches “kernel has IPv6” more than “this image has no IPv6 WAN”. On typical OpenWrt with `CONFIG_IPV6`, `lo` has `::1`, so the IPv4-only enable path mainly fires when IPv6 is compiled out or disabled with `ipv6.disable=1`.

Please spell that product rule in the helper comment / FAQ (and, if you still want WAN-only IPv6, this is the wrong file). Issue #371 left this as an open review point to validate on supported images.

docs/developer/security-review.md
```diff
@@ -141,6 +142,7 @@ should carry a note saying what would raise it.
...
 | `resolve` without `jshn` / with malformed input returns `error:jshn_missing` / `error:invalid_input` | `host` | same file `testResolveJshnMissing`; `invalid_input` runs only where `jshn` exists (skips on stock hosts). UI keeps the full resolve reply (no `expect` unwrap) so `disabled:load` / `error` siblings reach the view (#306 Layer 2). |
 | `logging_status` always returns the full 9-key shape; failures travel as `blockers`/`warnings` with `ready:false`, never a silent empty object | `host` | same file `testLoggingStatusNeverSilent` |
+| `nf_log` readiness is backend-aware and family-aware: empty/`NONE` selectors block a present family, while an independently unavailable IPv6 family is effective-ready | `host` | `tests/fwlive-logging.test.sh` #371 backend matrix, `/proc/net/if_inet6` fixtures, status and enable gates |
```
lucas-albers-lz4: **Nit — header still says “54 controls in force”** after this new row. Bump the count (or stop treating the banner number as a live total).

The proof class is also a bit strong until the enable tests actually assert `ok:true` on a stubbed lock rather than “not `nf_log_missing`”.

lucas-albers-lz4 (commented):
## Review (Grok) — #371 nf_log family-aware readiness

The production change matches the defect contract: empty/missing backends fail closed, `NONE` is case-insensitive, IPv6 availability is independent of `nf_log/10`, and status / blockers / enable share `compute_nf_log_state`. LuCI still collapses `nf_log_ipv4_missing` / `nf_log_ipv6_missing` into the existing generic missing-module copy. CI is green.

Two test gaps keep the enable gate from being proven. One probe semantic from #371’s open review points is still implicit.

### Findings

**CONFIRMED — IPv4-only enable is not proven, and the new test can touch `/etc/fwlive`.**
`tests/fwlive-logging.test.sh` calls real `enable_wan_logging` without stubbing `acquire_wan_log_lock` / `release_wan_log_lock` (every other enable case in this file does) and only asserts the reply is *not* `nf_log_missing`. On a root CI runner that can create `/etc/fwlive/logging.lock`; on a non-root host it returns `lock_failed` and still passes. #371 asked to prove IPv4-only enable succeeds.

**CONFIRMED — #371 family matrix is incomplete.**
The backend loop only asserts `read_nf_log_backend`. Status/blockers/ready/enable are covered for IPv4-ready + IPv6-absent and for present IPv6 + `NONE`, but not for “both families present and both loggers ready ⇒ ready”, and not for a present IPv4 family with `NONE`/empty/missing through the status/enable path.

**Open — `/proc/net/if_inet6` treats loopback `::1` as IPv6 present.**
Any non-empty `if_inet6` (including `lo`) requires `kmod-nf-log-ipv6`. That is a reasonable “IPv6 stack present” rule, but it is not an IPv4-only WAN probe: stock OpenWrt with IPv6 in the kernel almost always has `::1`. The IPv4-only bypass therefore mainly helps images with IPv6 compiled out or `ipv6.disable=1`. #371 asked to validate this on supported images and document the chosen signal.

**Nit — ledger header still says 54 controls** after adding the new `nf_log` readiness row.

tests/fwlive-logging.test.sh
lucas-albers-lz4: **FOLD** in `df25968`.

`enable_wan_logging` now runs under stubbed `acquire_wan_log_lock` / `release_wan_log_lock` with `FWLIVE_WAN_LOG_LOCK_FILE` inside the temp dir. IPv4-only enable asserts `"ok":true` (and rejects `nf_log_missing`). Present-IPv6 `NONE` still asserts `nf_log_missing` on the same stubs so a gate regression cannot mkdir `/etc/fwlive`.

tests/fwlive-logging.test.sh
lucas-albers-lz4: **FOLD** in `df25968`.

IPv4 missing/empty/`NONE`/`none` now each drive status (`nf_log_ipv4:false`, `ready:false`, `nf_log_ipv4_missing`) and `enable` → `nf_log_missing`. Added the dual-ready cell (both families present, both backends assigned → `ready:true` and `"ok":true`) next to the existing present-IPv6 `NONE` blocker.

openwrt-feed/luci-app-fwlive/root/usr/libexec/fwlive-logging.sh
lucas-albers-lz4: **FOLD** in `df25968` — kept the `if_inet6` content probe and documented it as an IPv6-stack signal, not a WAN-address probe.

Helper comments, FAQ, and `docs/fwlive-nft-logging.md` now say: missing/empty ⇒ stack absent (compiled out or `ipv6.disable=1`); any content, including `lo ::1`, ⇒ stack present and `kmod-nf-log-ipv6` required. Stock OpenWrt with `CONFIG_IPV6` still has loopback IPv6 on an IPv4-only WAN.

docs/developer/security-review.md
lucas-albers-lz4: **FOLD** in `df25968`. Header is 55 controls. The #371 row and method note now cite the stubbed-lock `"ok":true` enable assertions plus the dual-ready and IPv4-`NONE` cells.

lucas-albers-lz4:
## Independent review — verdict: APPROVE (draft; head verified: `6ef659e`)

Checked against the #371 design contract, with all gates run locally at the PR head (isolated worktree, `origin/master...pr-382` = exactly the 5 files GitHub reports).

**Contract compliance — all truth-table rows verified in code and by re-running the fixtures:**

| family | nf_log value | #371 requires | implementation |
|---|---|---|---|
| absent (IPv6, empty/missing `if_inet6`) | any | not required → passes | `nf_log_family_available ipv6 \|\| return 0` ✅ |
| present | missing/empty | blocks | `[ -n "$val" ] || return 1` ✅ |
| present | `NONE`/`none`/mixed case | blocks | `[Nn][Oo][Nn][Ee]` case pattern ✅ |
| present | real backend | passes | ✅ |

- **One computed state:** `compute_nf_log_state` runs once per entry point (`build_logging_status_json`, `enable_wan_logging`); `collect_logging_blockers` consumes the cached result (`NF_LOG_STATE_COMPUTED`) instead of re-reading the files — the duplicate IPv6 check is gone, and status fields / `ready` / blockers / enable gate cannot disagree. rpcd spawns a fresh process per call, so the cache can't go stale across RPCs.
- **Effective-readiness semantics** chosen for `nf_log_ipv4`/`nf_log_ipv6` (issue's preferred contract), and it *is* documented (security-review.md delta + FAQ + nft-logging.md) and tested (IPv4-only fixture asserts `nf_log_ipv6:true, ready:true`, no blocker). LuCI side needs no change: `logging.js` consumes blockers only, and an absent IPv6 family correctly produces no `nf_log_ipv6_missing` → no false "missing kernel log modules" toolbar text.
- **Generic `nf_log_missing` enable error preserved** (no API change), ACLs untouched, security-review.md updated in-PR per AGENTS.md.
- **Tests:** `tests/fwlive-logging.test.sh` — all three #371 sections pass locally. `./scripts/fwlive-shellcheck.sh` clean. Fixture seams are properly restored (save/unset/`trap` cycle is correct; the `FWLIVE_IPV6_AVAILABLE_PATH` re-check inside `nf_log_family_available` is what makes the post-source test override work).
- **CI:** all 12 checks green at this head.

### 🟡 Minor — `if_inet6` counts loopback-only as "IPv6 present"

The probe is `grep -q '[^[:space:]]'` on `/proc/net/if_inet6` — and the test fixture's "present" case is `…0001 01 80 lo`, i.e. loopback. On an image where the kernel has IPv6 but the admin disabled it on all interfaces (or `net.ipv6.conf.all.disable_ipv6=1`), `if_inet6` still exists with the `lo` entry, so IPv6 counts as *present* and a `NONE` selector at `nf_log/10` **blocks IPv4-only enablement**. That may be defensible (fw4 creates v6 rules whenever the kernel supports IPv6), but #371 explicitly asks to "validate the exact semantics on supported OpenWrt images and document the chosen signal" — the docs currently say "An absent /proc/net/if_inet6 means this image has no IPv6 interfaces", which overstates the probe: *loopback-only also counts as present*. Either document that semantics choice or exclude `lo`-only content. Worth a live-device check on an IPv6-disabled image before undrafting.

### 🟢 Nit — no whitespace trim before the NONE comparison

`val=$(cat …)` strips trailing newlines but not spaces. A selector containing `NONE ` (trailing space) would fail the case match and be treated as a real backend — fail-open, the exact defect class this PR fixes. Kernel sysctl output doesn't do this today, but this repo has been bitten by trailing-space key mismatches before (the fw4 `parseRuleHint` token saga). One `val="${val%%[[:space:]]}"`-style trim (or `case $val in *[![:space:]]*)`… ) would harden it cheaply.

### ℹ️ Environmental note (not a PR finding)

On my review host `/bin/sh` is dash, so the pre-existing `rpcd __selftest` section (`export -n`, bash-only) fails there — identically on stock master, and `busybox sh` passes the selftest on the PR head. CI is green. Unrelated to this diff.

**Bottom line:** the fail-open NONE defect is fixed, the family model matches #371's contract, and the state is genuinely computed once. The two notes above are hardening/documentation items, not blockers.

lucas-albers-lz4:
Review follow-up: the four inline findings are addressed and marked resolved. `df25968` added the lock-safe IPv4-only enable assertion, dual-ready and IPv4-`NONE` state cells, the explicit `if_inet6` IPv6-stack semantics, and the 55-control ledger count. `eed48b4` trims surrounding whitespace before interpreting `NONE`, with regression coverage for whitespace-padded selectors. Focused logging tests, ShellCheck, and the full `./scripts/fwlive-test.sh` suite pass.

lucas-albers-lz4:
## Triage of lasting findings (independent review vs HEAD `eed48b4`)

The independent review verified `6ef659e`. Current head is `eed48b4` (`df25968` + selector trim). Ground-truth of the two leftover notes:

| Label | Finding | Verdict at `eed48b4` |
| --- | --- | --- |
| **FOLD** | 🟡 `if_inet6` counts loopback-only as IPv6 present; docs overstated “no IPv6 interfaces” | Documentation gap is closed. Helper comments, FAQ, and `docs/fwlive-nft-logging.md` now say missing/empty `if_inet6` ⇒ stack absent (compiled out or `ipv6.disable=1`); any content, including `lo ::1`, ⇒ stack present and the IPv6 logger is required. That is an IPv6-stack probe, not a WAN-address probe. The `lo` fixture remains the explicit “present” cell. |
| **DISMISS** (as a code defect) | Same note: `net.ipv6.conf.all.disable_ipv6=1` with leftover `lo` would still require `nf_log/10` | Intended. fw4 still has an IPv6 family when the kernel stack is present. Excluding `lo` would be a different product rule than #371’s independent stack probe. |
| **FOLD** | 🟢 no whitespace trim before `NONE` | `read_nf_log_backend` now strips surrounding whitespace before the sentinel match. `tests/fwlive-logging.test.sh` covers ` NONE \t\n` on both family paths. |
| **DISMISS** | dash `export -n` selftest | Reviewer’s own environmental note; identical on master; not this diff. |

**Still open, not a merge blocker:** a live-device check that `ipv6.disable=1` / IPv6-compiled-out images actually have missing or empty `if_inet6`. Fixtures cover the contract; #371’s “validate on supported OpenWrt images” is still a lab residual, same as the independent review’s undraft note.

No remaining `CONFIRMED` code findings from that review at this head.

lucas-albers-lz4:
Follow-up for the leftover live-device check: #384. Product rule stays as documented (stack probe, including `lo ::1`). That lab issue does not block this PR.

PR #383, opened 2026-09-19
fix: resolve WAN zones by name or network

### LINKED ISSUE #373: WAN-zone lookup: resolve by name-or-network union so renamed zones work
Source of truth for the contract: #370 R1 is the defect fix; this issue owns the zone-identity behavior change (R7). Split from #371 deliberately: different function, different contract.

## Problem

`find_wan_zone_section` (`openwrt-feed/luci-app-fwlive/root/usr/libexec/fwlive-logging.sh` L173-191) matches the literal `name='wan'` only, then verifies `type == zone`. Any renamed WAN zone (`internet`, `WAN`, `wwan`) covering the `wan`/`wan6` networks resolves to empty, and `enable_wan_logging` / `logging_status` report `no_wan_zone` although a WAN zone is present.

## Decision (from #370 R7)

Union rule with implicit-network defaulting. A firewall section is the WAN zone when `type == zone` AND either holds:

1. `name == 'wan'` (exact, case-sensitive), or
2. its effective networks intersect `{wan, wan6}`, where effective networks = the explicit `network` list when present, else `[name]` (OpenWrt defaults an omitted `network` to the zone name).

Deliberate non-goals: no case-folding of names (UCI is case-sensitive; `WAN` over stock `wan` networks already resolves via rule 2). Full rename (zone `internet` over network `mywan`) stays `no_wan_zone` — undetectable without a user knob; the fix is a diagnostic that lists zones, not guessing. First match in config order wins (current behavior, keep).

## Scope

- Rewrite `find_wan_zone_section` per the union rule. Per candidate zone, add `uci -q get firewall.$sec.network` alongside the existing type probe; `uci get` normalizes `list network` lines and scalar `option network` into one space-separated string, so token-split covers both shapes. Missing/empty falls back to `[name]`.
- Check 25.12-era configs for `device`/`subnet`-scoped zones; include `device` tokens in the match set if they appear in the wild, else record investigated-and-dropped.
- Align ancillary `name='wan'` greps (`scripts/qemu-reset-wan-logging.sh`, uninstall smokes) or record why each stays.
- Improve the `no_wan_zone` diagnostic to include discovered zone names so a full rename is actionable.
- Same code path serves 21.02 (fw3) through 25.12 (fw4); zone `name`/`network` shapes are identical, so one implementation covers the ipk/apk split.

## Test matrix (host-stub, `tests/fwlive-logging.test.sh`)

Existing tests stay green (named, anonymous, non-zone skip, get-miss continue). Add:

- renamed zone `name='internet'` + `network='wan wan6'` → resolves (scalar form)
- same via multiple `list network` lines → resolves (list form)
- `name='WAN'` + `network='wan'` → resolves via membership, not name
- `name='wan'` + `network='lan'` → still resolves via name
- `name='wan'` + no `network` → resolves (existing behavior, now via defaulting)
- no match → empty, downstream `no_wan_zone` (enable/status error shape already covered)
- duplicate matches → first in config order

## Acceptance criteria

- [ ] Union rule implemented in `find_wan_zone_section`; strict membership-only and name-only both rejected by fixtures above
- [ ] Full matrix above passing under the required suite
- [ ] Ancillary greps aligned or explicitly dispositioned
- [ ] `no_wan_zone` diagnostic lists discovered zones

Refs #370 (R7 union decision).

lucas-albers-lz4:
Closes #373

## Summary
- Resolve the first firewall zone matching exact name=wan or effective network membership in wan/wan6.
- Treat omitted network options as the zone name, preserving existing name-only behavior and config order.
- Include discovered zone names in JSON-escaped no_wan_zone diagnostics.
- Align documentation and QEMU helpers with the shared logging_status wan_zone result.
- Record the 25.12 device/subnet investigation; no such firewall-zone shapes were found.

## Validation
- bash tests/fwlive-logging.test.sh
- ./scripts/fwlive-shellcheck.sh
- ./scripts/fwlive-test.sh

All pass.


lucas-albers-lz4 (commented):
Reviewed in an isolated worktree against `master`. The union rule itself matches #373 (exact `name=wan` **or** effective `wan`/`wan6` membership, omitted network defaults to the zone name, first match in config order, JSON-escaped candidates on enable/disable). Host logging tests, lock tests, and shellcheck pass here, but two of those greens are false and the user-facing diagnostic never reaches LuCI.

## Findings

1. **High — lock concurrency test is a false green.** `tests/fwlive-logging-lock.test.sh` Part B still stubs `uci show` as `firewall.@zone[0].name='wan'` with no `=zone` line. The new finder only iterates `firewall.<id>=zone`, so enable/disable return `no_wan_zone` and never touch the log bit. The suite still passes: unchanged is treated as a valid serial outcome, and `check_json` accepts any `{"ok":false,` payload. Reproduced: enable/disable both returned `no_wan_zone` and left `log=2` unchanged. Update the stub (`=zone` plus name/type/network `get`) and assert that at least one child is not `no_wan_zone`.

2. **Medium — `wan_zone_candidates` never reaches the UI path #373 called actionable.** Candidates are only on enable/disable error JSON. `logging_status` still emits `blockers: ["no_wan_zone"]` with no candidate list; LuCI hides the enable CTA and still shows generic “No WAN zone found”; `failureNotice` maps unknown errors to “Could not enable logging.” Put candidates on `logging_status` and render them with `textContent`.

3. **Medium — older lookup tests no longer hit the branches they name.** The finder now starts from `=zone` lines. `firewall.fwd=forwarding` and `firewall.gone.name='wan'` (no `=zone`) never enter the loop, so “skip non-zone” and “uci get miss continue” pass for the wrong reason.

4. **Low — screenshot helper can `uci set` with an empty zone id.** `zid=$(ubus … | jsonfilter …)` is used without a non-empty guard; `uci set firewall.$zid.log=…` becomes `firewall.log=…` if ubus/jsonfilter fails. Uninstall/security smokes already die on empty.

5. **Low — omitted `option name` on a named `wan` section.** Network defaulting uses `$_name`, not the section id. `config zone 'wan'` with neither `option name` nor `option network` stays `no_wan_zone`. Stock OpenWrt sets `option name 'wan'`; fall back to `$zone` if you want section-name-as-zone-name.

## Test matrix vs #373

Covered: renamed + `network=wan/wan6`, `WAN`+`wan`, `name=wan`+`network=lan`, omitted network on `name=wan`, first-match order, full-rename `no_wan_zone` with escaped quotes.

Gaps: lock stub (1); `logging_status` no-match shape; zone named `wan6` with omitted network (should match via defaulting); optional split of scalar vs list fixtures.

The 25.12 device/subnet check as investigated-and-dropped matches this tree (no such firewall-zone fixtures).

Not merge-ready until (1) and (2) are addressed.

openwrt-feed/luci-app-fwlive/root/usr/libexec/fwlive-logging.sh
```diff
@@ -170,26 +172,75 @@ release_wan_log_lock() {
...
+}
+
+wan_zone_diagnostic_json() {
+	printf '[%s]' "${WAN_ZONE_DIAGNOSTIC_JSON:-}"
+}
+
+no_wan_zone_error_json() {
+	_candidates=$(wan_zone_diagnostic_json)
+	printf '{"ok":false,"changed":false,"wan_zone":null,"error":"no_wan_zone","wan_zone_candidates":%s}' \
+		"$_candidates"
```
lucas-albers-lz4, line 255: **Medium:** `wan_zone_candidates` is only attached here, on enable/disable error JSON.

#373 asked for discovered zone names so a full rename is actionable. LuCI never sees this payload:
- `logging_status` still reports `blockers: ["no_wan_zone"]` with no candidate list
- the empty state stays generic “No WAN zone found” (`logging.js`)
- the enable CTA is hidden when that blocker is present
- if enable did run, `failureNotice` maps unknown errors to “Could not enable logging.”

SSH `ubus call fwlive enable_wan_logging` is the only place the list appears. Add candidates to `build_logging_status_json` (and pin the key in `testLoggingStatusNeverSilent`) and render them with `textContent` in the no-zone empty state.

openwrt-feed/luci-app-fwlive/root/usr/libexec/fwlive-logging.sh
```diff
@@ -170,26 +172,75 @@ release_wan_log_lock() {
...
+		WAN_ZONE_DIAGNOSTIC_JSON="${WAN_ZONE_DIAGNOSTIC_JSON}\"${_esc}\""
+
 		# uci -q get exits 1 on a missing section. Capture with || true so
 		# set -e cannot abort inside "$(…)" before || continue.
 		_type=$(uci -q get "firewall.${zone}" 2>/dev/null || true)
 		[ "$_type" = "zone" ] || continue
-		printf '%s' "$zone"
-		return 0
+		_network=$(uci -q get "firewall.${zone}.network" 2>/dev/null || true)
+		[ -n "$_network" ] || _network="$_name"
```
lucas-albers-lz4, line 229: **Low:** omitted-network defaulting uses `$_name`, not the section id. `config zone 'wan'` with neither `option name` nor `option network` stays `no_wan_zone` even though OpenWrt treats the section name as the zone name.

Stock images always set `option name 'wan'`, so this is an edge case. If you want that OpenWrt behavior, fall back to `$zone` when name is empty (the diagnostic label already does: `_zone_label="$_name" || "$zone"`).

openwrt-feed/luci-app-fwlive/root/usr/libexec/fwlive-logging.sh
```diff
@@ -491,7 +543,8 @@ json_null_or_string() {
 }
 
 build_logging_status_json() {
-	zone=$(find_wan_zone_section)
+	find_wan_zone_section_state
+	zone=$WAN_ZONE_FOUND
```
lucas-albers-lz4, line 606: **Medium:** this status path records `WAN_ZONE_DIAGNOSTIC_JSON` during lookup, then drops it. A no-zone status still has only `blockers: ["no_wan_zone"]`.

This is the reply the UI actually reads. Without `wan_zone_candidates` here, the enable/disable diagnostic never reaches the empty state.

scripts/capture-fwlive-screenshots.mjs
```diff
@@ -61,7 +61,7 @@ async function ensureWanLoggingOff(page) {
 	/* Guest may leave /tmp/.uci staged (@zone vs cfgXXXX) so the UI toggle
 	 * returns firewall_changes_pending. Force off via UCI for shot 1. */
 	if (await page.locator('#fwlive-logging-bar button', { hasText: 'WAN logging on' }).count()) {
-		guestSsh('rm -rf /tmp/.uci; mkdir -m 0700 /tmp/.uci; zid=$(uci show firewall | sed -n "s/^firewall\\.\\([^.]*\\)\\.name=.wan.$/\\1/p" | head -1); uci set firewall.$zid.log=0; uci commit firewall; /etc/init.d/firewall reload; rm -rf /tmp/.uci; mkdir -m 0700 /tmp/.uci');
+		guestSsh("rm -rf /tmp/.uci; mkdir -m 0700 /tmp/.uci; zid=$(ubus call fwlive logging_status 2>/dev/null | jsonfilter -e '$.wan_zone' 2>/dev/null); uci set firewall.$zid.log=0; uci commit firewall; /etc/init.d/firewall reload; rm -rf /tmp/.uci; mkdir -m 0700 /tmp/.uci");
```
lucas-albers-lz4: **Low:** `zid` is not checked before `uci set firewall.$zid.log=0`. If `logging_status` or jsonfilter fails, this becomes `uci set firewall.log=0`.

`qemu-logging-uninstall-smoke.sh` / `qemu-security-gaps-smoke.sh` already `die` on empty. Guard both screenshot fallbacks with `[ -n "$zid" ]` (same at L145).

tests/fwlive-logging.test.sh
```diff
@@ -227,6 +227,120 @@ got=$(find_wan_zone_section)
...
+case "$out" in
+	*'"error":"no_wan_zone"'*'"wan_zone_candidates":["internet\"edge"]'*) ;;
+	*) die "no_wan_zone response must list discovered zones, got: $out" ;;
+esac
+out=$(disable_wan_logging)
+case "$out" in
+	*'"error":"no_wan_zone"'*'"wan_zone_candidates":["internet\"edge"]'*) ;;
+	*) die "disable no_wan_zone response must list discovered zones, got: $out" ;;
+esac
+ok "no_wan_zone diagnostic lists fully renamed zones"
```
lucas-albers-lz4, line 569: New matrix looks right for the union rule. Two related test gaps are **not** in this hunk:

**High — lock test false green (file not in this PR):** `tests/fwlive-logging-lock.test.sh` Part B still stubs `uci show` as `firewall.@zone[0].name='wan'` with no `=zone` line. New finder yields no zones; enable/disable return `no_wan_zone` and never toggle the log bit. The 32-trial sweep still passes because unchanged is a valid serial outcome and `check_json` accepts `{"ok":false,`. Reproduced against this commit: log stayed `2`. Add `=zone` plus name/type/network `get`, and reject `no_wan_zone` as a success.

**Medium — tests immediately above this block are false green.** Finder now iterates only `=zone` lines, so:
- “skip non-zone `name=wan`” never sees `firewall.fwd=forwarding`
- “uci get miss continue” never sees `firewall.gone.name='wan'` (no `=zone`)

Both still resolve `@zone[0]`. If those branches still matter, the fixtures need `=zone` plus a failing type `get`.

Also missing vs #373: `logging_status` no-match shape with candidates; zone named `wan6` with omitted network (should match via defaulting).

lucas-albers-lz4:
Review follow-up: all five inline findings are addressed and marked resolved in `3702387`. `logging_status` now carries `wan_zone_candidates`, and the LuCI no-zone state renders candidate names through text-node children with XSS coverage. Section-id defaulting handles omitted `name`/`network`; screenshot fallbacks reject empty zone ids; the lock-test UCI fixture now declares a real zone and rejects `no_wan_zone` as success. Finder coverage now includes `wan6` defaulting, no-zone status shape, and corrected non-zone/get-miss paths. The full `./scripts/fwlive-test.sh` suite passes.

openwrt-feed/luci-app-fwlive/root/usr/libexec/fwlive-logging.sh
lucas-albers-lz4: **FOLD** on `3702387`.

`logging_status` now emits `wan_zone_candidates`, LuCI empty state renders them as `E('code', {}, [joined])` children (text nodes), and the e-harness pins a hostile candidate with zero `innerHTML` writes. `testLoggingStatusNeverSilent` requires the array key.

Residual (non-blocking): the toolbar no-zone span is still the generic “WAN logging unavailable: no WAN zone”; the empty state is the path #373 called actionable.

openwrt-feed/luci-app-fwlive/root/usr/libexec/fwlive-logging.sh
lucas-albers-lz4: **FOLD** on `3702387`.

`build_logging_status_json` now includes `wan_zone_candidates` from the same lookup. Host fixture asserts `wan_zone:null` plus escaped `["internet\"edge"]`.

openwrt-feed/luci-app-fwlive/root/usr/libexec/fwlive-logging.sh
lucas-albers-lz4: **FOLD** on `3702387`.

Finder now does `[ -n "$_name" ] || _name="$zone"` before network defaulting, so a named `wan` section with neither `option name` nor `option network` matches via the section id.

Residual: no host fixture for that omitted-name shape (code path is there; `wan6` omitted-network is covered).

scripts/capture-fwlive-screenshots.mjs
lucas-albers-lz4: **FOLD** on `3702387`.

Both screenshot fallbacks now `[ -n "$zid" ] || exit 1` before `uci set "firewall.$zid.log=…"`.

tests/fwlive-logging.test.sh
lucas-albers-lz4: **FOLD** on `3702387`.

Lock Part B stub now emits `=zone` plus name/type/network `get`, and `check_json` rejects `error:no_wan_zone`. Reproduced against this commit: lookup returns `wan_zone":"@zone[0]"` (previously `no_wan_zone`). Skip-non-zone is now show `=zone` + type get `forwarding`; get-miss is `gone=zone` with failing type get. Also covered: `wan6` omitted-network defaulting and `logging_status` no-match candidates.

32-trial lock sweep, logging tests, e-harness, and rpcd security pass on this commit.

lucas-albers-lz4:
Re-review against `c777651`:

- The five prior findings remain fixed and all five review threads are resolved.
- New CI finding: the toolbar follow-up in `e447878` left `logging.js` out of Prettier format, causing the `test` workflow to fail before the functional suite ran.
- Fix: formatted the candidate renderer in `c777651` and pushed it.
- Verification: `./scripts/fwlive-test.sh` passes locally, including the WAN lookup, toolbar/empty-state text-node XSS coverage, lock race, and package checks.

No new behavioral or security findings. GitHub checks need to complete for the new commit; CodeRabbit has only posted its draft-PR notice, not a substantive review.

lucas-albers-lz4:
Post-#382 merge update: master now contains #382 (`fdf462f`). This branch was updated with that merge and the two overlapping files were resolved to preserve both nf_log readiness and WAN-zone diagnostics (`f928e68`). The full `./scripts/fwlive-test.sh` suite passes on the combined branch. GitHub checks are rerunning; review threads remain resolved.

Issue #384, opened 2026-09-19
lab: validate if_inet6 IPv6-stack probe on supported OpenWrt images


lucas-albers-lz4:
## Why this issue exists

#371 / #382 settled the product rule: `/proc/net/if_inet6` is an **IPv6-stack** probe, not a WAN-address probe.

- missing or empty file → stack absent (compiled out or `ipv6.disable=1`) → IPv6 logger not required
- any content, including loopback `::1` → stack present → `kmod-nf-log-ipv6` required

Host fixtures already cover missing / empty / `lo`. This issue is only the live-device check #371 asked for before treating those semantics as proven on supported OpenWrt images. It is **not** a request to exclude `lo` or to delay merging #382.

## What to record (per image)

On each guest, capture:

```sh
cat /proc/net/if_inet6
cat /proc/sys/net/netfilter/nf_log/2
cat /proc/sys/net/netfilter/nf_log/10
ubus call fwlive logging_status
```

Expected:

| Image condition | `if_inet6` | IPv6 logger required |
| --- | --- | --- |
| stock 24.10 / 25.12 with `CONFIG_IPV6` | at least `lo ::1` | yes |
| IPv4-only WAN, IPv6 stack still on | still `lo ::1` | yes |
| `ipv6.disable=1` boot or IPv6 compiled out | missing or empty | no |

`net.ipv6.conf.all.disable_ipv6=1` with leftover `lo` is expected to keep the IPv6 logger required. That is not a probe bug.

## Acceptance

- [ ] Stock QEMU guest (`docs/developer/qemu-lab.md`, current 24.10 and 25.12 pins) shows populated `if_inet6` including `lo`, and `logging_status` requires the IPv6 backend when `nf_log/10` is `NONE`.
- [ ] At least one no-stack case (boot `ipv6.disable=1` or an IPv6-less image) shows missing/empty `if_inet6`, and IPv4-ready enable is allowed.
- [ ] Results folded into `docs/developer/security-review.md` (proof class can stay `host` for fixtures; lab note records the device check).
- [ ] Reopen the probe design only if a supported image contradicts the table (for example `ipv6.disable=1` still has a populated `if_inet6`).

## Non-goals

- Changing the probe to ignore loopback.
- Treating IPv4-only WAN as IPv6-absent while the kernel stack is present.
- Blocking #382.

Refs #371

PR #387, opened 2026-09-19
feat(logging): warn on legacy iptables tables

lucas-albers-lz4:
## Summary

- detect registered legacy IPv4/IPv6 iptables tables by reading the procfs table-name files
- add `legacy_iptables_detected` to the informational `logging_status.warnings` array
- keep the warning diagnostic-only: it does not affect readiness, blockers, or the enable gate
- keep the warning ubus-only for this slice and document that choice in the security review
- add fixture-based coverage for missing, empty, IPv4-populated, and IPv6-populated table-name files

## Validation

- `bash tests/fwlive-logging.test.sh`
- `./scripts/fwlive-shellcheck.sh`
- `./scripts/fwlive-test.sh`

Refs #378

## Summary by CodeRabbit

- **New Features**
  - Added a diagnostic warning when legacy IPv4 or IPv6 iptables tables are detected.
  - This warning does not affect readiness, blockers, or enablement.

- **Documentation**
  - Documented the `legacy_iptables_detected` warning and its diagnostic-only behavior.

- **Tests**
  - Added coverage for missing, empty, and populated legacy iptables table files.


tests/fwlive-logging.test.sh
```diff
@@ -42,6 +42,42 @@ case "$out" in
...
+esac
+: >"$LEGACY_IPV4"
+printf 'filter\n' >"$LEGACY_IPV6"
+out=$(build_logging_status_json)
+case "$out" in
+	*'legacy_iptables_detected'*) ;;
+	*) die "IPv6 legacy table must warn: $out" ;;
+esac
+rm -rf "$LEGACY_WORK"
+unset FWLIVE_IP_TABLES_NAMES_PATH FWLIVE_IP6_TABLES_NAMES_PATH
```
lucas-albers-lz4: **Host isolation leak.** After this `unset`, later `build_logging_status_json` calls fall back to `/proc/net/ip_tables_names` and `/proc/net/ip6_tables_names`.

The existing exact pin below then fails if those files are readable and populated:

```sh
*'"warnings":["timeout_missing"]'*)
```

becomes `"warnings":["timeout_missing","legacy_iptables_detected"]`.

Reproduced with a readable fixture containing `filter\n`. CI can still pass because those proc files are often `-r--r-----` on Ubuntu runners, so `[ -r ]` is false for the test user. That is latent, not proof of host independence. #378 asked not to depend on host kernel state.

Leave empty fixture files exported for the rest of the file (same idea as the nf_log workdir), or re-pin empty paths around the `timeout_missing` assert.

tests/fwlive-logging.test.sh
```diff
@@ -42,6 +42,42 @@ case "$out" in
...
+case "$out" in
+	*'legacy_iptables_detected'*) ;;
+	*) die "IPv4 legacy table must warn: $out" ;;
+esac
+: >"$LEGACY_IPV4"
+printf 'filter\n' >"$LEGACY_IPV6"
+out=$(build_logging_status_json)
+case "$out" in
+	*'legacy_iptables_detected'*) ;;
+	*) die "IPv6 legacy table must warn: $out" ;;
```
lucas-albers-lz4: These cases only search for `legacy_iptables_detected` anywhere in the JSON. That still passes if the token is appended to `blockers`, or if `ready` flips — the core #378 contract.

Pin field location the way `timeout_missing` already does: present in `warnings`, absent from `blockers`. Add the missing **both-populated** fixture from the issue (production already dedupes to one warning; it is untested).

docs/developer/security-review.md
```diff
@@ -564,3 +564,18 @@ WAN does not make the stack absent. `nf_log/10` staying populated under
...
+
+**Scope.** `fwlive-logging.sh` reads the network namespace's IPv4 and IPv6
+legacy table-name procfs files and appends `legacy_iptables_detected` to
+`logging_status.warnings` when either contains a table name. The warning is
+diagnostic-only: it does not affect `ready`, `blockers`, or the enable gate,
+and remains ubus-only until a user-facing remediation flow is designed.
+
+**Method.** Host fixtures cover missing, empty, IPv4-populated, and
+IPv6-populated table-name files through `FWLIVE_IP_TABLES_NAMES_PATH` and
+`FWLIVE_IP6_TABLES_NAMES_PATH`; no host kernel state or privilege is required.
```
lucas-albers-lz4: This is overstated once the test unsets `FWLIVE_IP_*_TABLES_NAMES_PATH`. The rest of `tests/fwlive-logging.test.sh` can still observe host procfs (and the exact `timeout_missing` warnings array depends on it).

After the test isolation fix, this sentence is accurate. Until then it should not claim the full suite is fixture-only.

lucas-albers-lz4 (commented):
## Review

Production change is a correct Phase 1 slice of #378: content-probe both table-name files, `legacy_iptables_detected` on `logging_status.warnings` only, enable/UI still ignore this warning. Independently re-checked against the diff and a simulated populated proc file.

Please fix the two test gaps before merge. (Posted as COMMENT because GitHub forbids REQUEST_CHANGES on your own PR.) CI green is not enough here: on typical Ubuntu runners `/proc/net/ip_tables_names` is often root-only, so `[ -r ]` stays false and the suite can pass while still depending on host kernel/permissions.

### Blockers
1. **Host isolation leaks after the new fixtures** — unsetting the path overrides lets later tests, including the exact `timeout_missing` array pin, read real procfs.
2. **Diagnostic-only contract is not pinned** — substring search anywhere in the JSON would still pass if the token landed in `blockers`. Missing the both-populated fixture from #378.

### Nits
- Security-review **Current delta** header still describes #384/#371/#383.
- Ledger **Method** overstates fixture isolation once the env vars are unset.
- `[Unreleased]` changelog is empty for a new `logging_status` warning string.

Not issues: `grep -q '[^[:space:]]'` matches the existing `if_inet6` probe; `iptables-nft` should not populate these legacy files; `set -e` with `legacy_iptables_active && …` is safe.

lucas-albers-lz4:
## Review — warn on legacy iptables tables (`b621fd91`)

Independent pass over the diff, the consumers, and the acceptance criteria in #378. Verdict: **the shipped shell change is correct and the diagnostic-only contract holds; the test file does not yet meet two of #378's stated requirements, and two documentation statements are wrong.** Fixes are small and all sit in the same two files.

### What I verified as correct (no action)

| Claim in the PR | Evidence |
|---|---|
| Diagnostic-only: no effect on `ready`, `blockers`, or the enable gate | `ready` is computed from `zone`/`wan_log`/`nf4`/`nf6` only (`fwlive-logging.sh:647-648`); `blockers` comes from `collect_logging_blockers` (`:642`). The new token's only consumer is the `warnings` array (`:645`). |
| Detects registered tables, not a loaded module | Content probe (`[ -r ]` + `grep -q '[^[:space:]]'`), not module presence — and not `[ -s ]`, so the procfs `st_size == 0` caveat called out in #378 is honoured. |
| Both families read, production defaults intact | `${FWLIVE_IP_TABLES_NAMES_PATH:-/proc/net/ip_tables_names}`, `${FWLIVE_IP6_TABLES_NAMES_PATH:-/proc/net/ip6_tables_names}` (`:584-586`). |
| ubus-only, no DOM sink, no catalog churn | The view consults only `timeout_missing` (`view/status/fwlive.js:661-666`), so the new token is silently ignored in the UI. #378 only required the choice to be documented — it is, in `security-review.md`. |
| House conventions | Tabs throughout; `./scripts/fwlive-shellcheck.sh` → OK (baseline clean). No `CHANGELOG`/version bump needed — `CHANGELOG.md` is only touched by release commits. |

### Findings

**1. `tests/fwlive-logging.test.sh:78` — test isolation leak (fix; contradicts a #378 requirement).**
The new block unsets both path overrides, so every later `build_logging_status_json` call probes the real procfs. Reproduced: with a readable table-name file containing `filter`, the warnings array becomes `["timeout_missing","legacy_iptables_detected"]` and the exact pin at `:334` (`*'"warnings":["timeout_missing"]'*`) fails →

```
fwlive-logging test FAIL: expected timeout_missing in warnings, got: {"warnings":["timeout_missing","legacy_iptables_detected"],...}
```

It is latent here only because this host's `-r--r-----` 0-byte `/proc/net/ip_tables_names` files are empty, and because GitHub runners run the suite as a non-root user where `[ -r ]` is false. Any root/container/QEMU-style run on a host with a legacy table registered breaks it. #378 requires the host tests to "not depend on the host kernel or require a privileged test".
**Fix:** keep the empty fixture files exported for the rest of the file — the `NF_LOG_WORK` pattern at `:100-107` is the model — or re-pin empty paths around the `timeout_missing` assert.

**2. `tests/fwlive-logging.test.sh:45-76` — assertions do not pin field location (fix; the exact contract #378 states).**
All four cases only match `*'legacy_iptables_detected'*` anywhere in the JSON, so they still pass if the token is appended to `blockers`, or if `ready` flips — the one behaviour the warning "must not affect". The file already has the right shape three lines below, for `timeout_missing` (`:333-339`: present in `warnings`, absent from `blockers`). Copy that pattern; add a `ready` assertion while you are there.

**3. `tests/fwlive-logging.test.sh:45-76` — the both-populated fixture is missing (fix; named acceptance criterion).**
#378 asks for "missing, empty, IPv4-populated, IPv6-populated, **and both-populated** cases"; four of five are present. Production dedupes to a single token by construction (short-circuit on the first match — confirmed: with both files populated `legacy_iptables_active` returns 0 and one token is emitted), so this is a regression guard rather than a behaviour gap, but it is an explicit AC.

**4. `docs/developer/security-review.md:149` — "the full 9-key shape" is stale (fix, one word).**
`build_logging_status_json` always emits 10 keys (`fwlive-logging.sh:640-641`: `wan_zone`, `wan_zone_candidates`, `wan_log`, `wan_log_limit`, `nf_log_ipv4`, `nf_log_ipv6`, `ready`, `weak_device`, `blockers`, `warnings`), and `tests/fwlive-rpcd-security.test.js:512-523` pins the same 10. This PR is editing that exact row, so it should leave it accurate: **10-key shape**.

**5. `docs/developer/security-review.md` — "no host kernel state or privilege is required" (fix, tied to finding 1).**
Accurate for the four fixture cases in isolation, but not for the suite as it stands, because of finding 1. Fixing 1 makes the sentence true as written; otherwise scope it explicitly to the fixture probe.

**6. `fwlive-logging.sh:584-591` — duplicated procfs content probe (optional, same pass or defer).**
`nf_log_family_available` already encodes the identical "readable + any non-space byte" probe (`:520-522`). Extracting one helper (e.g. `procfs_entry_has_content <path>`) and using it in both keeps the "missing/empty ⇒ false, any content ⇒ true" semantics defined in one place — and removes the second leaked `_path`-style global (the file declares no `local`s anywhere, so a helper is the cheaper way to contain that).

**7. Proof class (optional, deferrable).**
The table row keeps proof class `host` via fixtures, while the section also asserts a device-level fact — that supported releases do not spuriously populate these files. The `nf_log` row closes the analogous gap with a QEMU lab note (24.10.8 / 25.12.5). A one-line lab check that `logging_status.warnings` stays empty on a clean 24.10/25.12 device would retire the assumption. Not a blocker for this slice.

**8. Informational, no change.** The probe reports tables *registered* in the netns, which is what #378 asked for. Worth knowing operationally: once any legacy tool touches a table — even a read-only `iptables-legacy -L` — that name persists in `ip_tables_names`, so the warning can outlive the condition it describes. That matches the issued wording, and it is diagnostic and ubus-only, so no change is recommended.

### Recommendation

Fix **1, 2, 3** (test file) and **4, 5** (docs) before merge — one commit covers all five, and 1 and 2/3 are the AC gaps rather than style. **6** fits the same pass if the helper is extracted while the file is open; **7** can ride a later slice.

### Local verification note

`bash tests/fwlive-logging.test.sh` does not go fully green on my machine: it dies in the rpcd `__selftest` (`rpcd/fwlive: 125: export: Illegal option -n`, `poll_lines_from_input: expected 2000 got 50`). The identical failure reproduces on `master` (`d44c0610e7`), so it is pre-existing and environment-specific, not attributable to this PR. The new #378 block and the `timeout_missing` pin both pass here.

Refs #378

docs/developer/security-review.md
```diff
@@ -146,7 +146,7 @@ should carry a note saying what would raise it.
...
 | `resolve` without `jshn` / with malformed input returns `error:jshn_missing` / `error:invalid_input` | `host` | same file `testResolveJshnMissing`; `invalid_input` runs only where `jshn` exists (skips on stock hosts). UI keeps the full resolve reply (no `expect` unwrap) so `disabled:load` / `error` siblings reach the view (#306 Layer 2). |
-| `logging_status` always returns the full 9-key shape; failures travel as `blockers`/`warnings` with `ready:false`, never a silent empty object | `host` | same file `testLoggingStatusNeverSilent` |
+| `logging_status` always returns the full 9-key shape; failures travel as `blockers`/`warnings` with `ready:false`, never a silent empty object | `host` | same file `testLoggingStatusNeverSilent`; #378 `legacy_iptables_detected` warning uses fixtureable procfs table-name probes and remains diagnostic-only |
```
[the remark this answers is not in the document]
lucas-albers-lz4: Fixed in ed323af. The row now says 10-key shape, matching `build_logging_status_json` and `testLoggingStatusNeverSilent`.

tests/fwlive-logging.test.sh
lucas-albers-lz4: Fixed in ed323af. Empty table-name fixtures stay exported for later `build_logging_status_json` calls, including the exact `timeout_missing` warnings pin.

tests/fwlive-logging.test.sh
lucas-albers-lz4: Fixed in ed323af. Cases now pin the token in `warnings`, absent from `blockers`, and `ready` unchanged vs the empty-file baseline. Added the both-populated fixture.

docs/developer/security-review.md
lucas-albers-lz4: Fixed in ed323af. Method now records that empty fixtures remain exported so later tests do not read host procfs, and that both-populated is covered.

lucas-albers-lz4:
Addressed the independent review in ed323af:

- **1–3 (tests):** empty table-name fixtures stay exported; assertions pin `warnings` vs `blockers` and unchanged `ready`; both-populated case added.
- **4–5 (docs):** `logging_status` row is 10-key; Method now matches the fixture isolation.

Deferred **6** (shared procfs helper) and **7** (QEMU empty-warning lab note) as optional follow-ups, not this slice.

`bash tests/fwlive-logging.test.sh` and `./scripts/fwlive-shellcheck.sh` passed here, including rpcd `__selftest`.

lucas-albers-lz4:
@coderabbitai review

Issue #393, opened 2026-09-19
test: cover hash restoration and strengthen chip sink assertions


lucas-albers-lz4:
Parent review: #370 (R3/G3)

Strengthen the LuCI filter/chip test contract. The current test checks only the outer container's HTML writes, and hash restoration through applyHash is not covered.

Acceptance criteria:
- recursively inspect sink records and assert attacker-controlled bytes never reach an HTML sink;
- cover negated q, src, and dst filters, contains text, and the not-equal marker using visible text assertions;
- exercise valid and malformed filter hashes through applyHash without throwing;
- cover hostile hash values as text nodes, not markup;
- document that renderer tests are complementary evidence and not standalone XSS proof.

lucas-albers-lz4:
## Review note — targets for the criteria

Verified against `master @ c6c8a90347`. Scope unchanged; two pointers so the criteria can be met without new machinery.

- **The recursive sink inspection already exists — reuse it.** `collectSubtreeInnerHTMLWrites` (`tests/fwlive-e-harness.test.js:135`) and `collectInnerHTMLWrites` (`tests/fwlive-modules-smoke.test.js:409`) both walk a subtree. The gap is narrow: the chip test inspects only the container it was handed (`tests/fwlive-proto-filter.test.js:91-94`), never the subtree `renderFilterChips` builds. Point at one of those helpers rather than adding a third copy.
- **Cite the `applyHash` location.** It is `view/status/fwlive.js:347`, invoked at `:2460`, so the "no throw on malformed" criterion has a named target.

Confirmed absent at master, matching this issue: any `applyHash` reference under `tests/` (hash coverage is `limit` / `poll` / `maxraw` preferences only, `tests/fwlive-view-fetch-budget.test.js`), and any negated `q` / `src` / `dst`, `contains`, or not-equal assertion in `fwlive-proto-filter.test.js`.

Refs #370

lucas-albers-lz4:
Recursive sink inspection now reuses one helper: `collectInnerHTMLWrites` / `collectSubtreeInnerHTMLWrites` exported from `tests/lib/luci-e-harness.js`. `applyHash` coverage is `view/status/fwlive.js` restore (valid, malformed `%`, hostile values as text nodes).

Refs #370

lucas-albers-lz4:
Resolved by merged PR #394: recursive HTML-sink assertions, hostile filter/chip text assertions, hash restoration cases, and renderer-test evidence limitations are covered.

Issue #392, opened 2026-09-19
test: pin fwlive ACL method ownership and installed session boundaries


lucas-albers-lz4:
Parent review: #370 (R4a/R4b)

Make the ACL contract exact rather than only checking method-set union and read/write disjointness.

Acceptance criteria:
- assert the exact read methods: rules, poll, resolve, and logging_status;
- assert the exact write methods: enable_wan_logging and disable_wan_logging;
- retain the invariant that sessions do not receive ubus log.read;
- record an installed authenticated-session check for a user with the grant and a user without it;
- keep the host test and installed evidence clearly separated.

lucas-albers-lz4:
Stacked closeout (draft, waiting human review):

- #408 docs nits on the #389 evidence artifact (dual `_all` IPK hashes, parallel-lab port 2224, v0.1.44 `luci.mk` ordering)
- #409 ledger closeout stacked on #408 (`Closes` keywords live there)

#209 and #307 are intentionally left open.

Issue #391, opened 2026-09-19
test: cover shell filter and toggle fail-closed error paths


lucas-albers-lz4:
Parent review: #370 (R5)

Add direct and integrated coverage for the remaining shell and toggle error paths.

Acceptance criteria:
- cover the real filter's separate mktemp failure path;
- cover symlink-directory rejection;
- cover integrated toggle replies for lock and snapshot failures;
- assert no partial staging or configuration mutation on failure;
- run relevant shell behavior under the supported dash and BusyBox test paths.

lucas-albers-lz4:
## Review note — name the targets in two criteria

Verified against `master @ c6c8a90347`. Scope unchanged; both criteria are right but ambiguous enough that an implementer could test the wrong guard.

- **"symlink-directory rejection" spans three scripts.** Name the one in scope: `fwlive-logging.sh:151-161` (the lock directory refuses a symlink, re-checked after `mkdir` for TOCTOU), `fwlive-adaptive-cap.sh:209` (refuses a symlinked state path), and `fwlive-log-filter.sh:45-53` (sticky-directory check plus its own `mktemp`). They are separate guards, not one.
- **"integrated toggle replies for lock and snapshot failures" — the tokens already exist**, so the criterion can be exact: `{"ok":false,…,"error":"lock_failed"}` and `{"ok":false,…,"error":"baseline_snapshot_failed"}`, emitted from `enable_wan_logging` (`fwlive-logging.sh:851` ff.) and mirrored in `disable_wan_logging` (`:913` ff.). Citing the tokens removes any guesswork about the reply shape.
- **Confirmed open, matching this issue:** the filter's separate `mktemp` path is `fwlive-log-filter.sh:53` (distinct from the sticky check above it at `:45-52`), and the lock helper's failure side is covered today only at helper level (`tests/fwlive-logging-lock.test.sh:82-107`, PATH-shadowed `flock` fails closed) — not through the toggle reply. No partial-staging assertion exists either.

Refs #370

lucas-albers-lz4:
Resolved by merged PR #394: shell filter mktemp/symlink fail-closed paths and integrated toggle lock/snapshot failure coverage were added and validated.

Issue #390, opened 2026-09-19
test: expand LuCI browser fixtures for real reply and failure contracts


lucas-albers-lz4:
Parent review: #370 (R2/R6)

Expand the mocked-browser and renderer evidence beyond benign IPv4 rows and the currently covered message cases.

Acceptance criteria:
- cover one-line, ANSI, and overlong message rendering;
- add IPv6 and hostile-row fixtures;
- exercise adaptive summary/degradation and warning rendering;
- cover storage failure and resolver-error reply paths;
- keep fixtures aligned with the production RPC reply shape;
- preserve text-node-only handling for untrusted values.

lucas-albers-lz4:
## R6 backend-label remainder

The 2026-09-19 dispositions review on #370 is right that `backend: 'iptables'` / `'unknown'` rendering was left out of this issue's criteria. Host coverage for those `backendDisplayLabel()` / `updateBackendUi()` branches is landing with the #370 host-coverage work (`tests/fwlive-view-logging-toggle.test.js`). This issue still owns the mocked-browser fixture half (IPv6 / hostile rows, summary/shed, storage failure, warnings, resolver errors).

Refs #370

lucas-albers-lz4:
Implemented and pushed the browser-fixture expansion in PR #399.

The mocked Playwright harness now covers IPv6, one-line/ANSI/overlong messages, hostile-row text-node safety, adaptive summary/shedding, warning rendering, storage failures, and resolver-error replies while keeping the production RPC reply shape.

Verification: `npm run test:view` and the full `./scripts/fwlive-test.sh` host suite pass.

lucas-albers-lz4:
Reviewer findings from merged PR #399 are addressed in follow-up PR #400: https://github.com/lucas-albers-lz4/fwlive/pull/400

Covered: self-contained Detail/one-line fixture setup, IPv6 assertions in rendered address cells, a real hostile <img> summary payload for text-node safety, and finally-based restoration for poll/resolver mocks. Focused browser smoke and full ./scripts/fwlive-test.sh pass.

PR #394, opened 2026-09-20
test: remaining #370 host coverage (chips, ACL, filter, table)

lucas-albers-lz4:
## Summary
- Host tests for the remaining #370 coverage items that do not need QEMU or a mocked browser: chip sink + `applyHash` (Refs #393), exact ACL read/write arrays (Refs #392 R4a), filter `mktemp`/`-L` and toggle `lock_failed`/`baseline_snapshot_failed` (Refs #391), table oneline/ANSI/overlong sinks plus iptables/unknown backend labels (Refs #390 R2/R6 host).
- Product fix in `chips.js`: negated `q`/`src`/`dst` chips now key off `spec.key` so the search field labeled `search` still renders as `contains`.
- Shared `collectInnerHTMLWrites` helper in the LuCI E harness; G4 `RESOLVE_MAX=32` is now greppable next to the existing 33-address cap.

Deferred on purpose: #389/#378 QEMU cells, #392 R4b installed sessions, #390 mocked-browser fixtures.

## Test plan
- [ ] `node tests/fwlive-chips-hash.test.js`
- [ ] `node tests/fwlive-proto-filter.test.js`
- [ ] `node tests/fwlive-modules-smoke.test.js`
- [ ] `node tests/fwlive-rpcd-security.test.js`
- [ ] `node tests/fwlive-shell-filter.test.js` (and `SH=busybox sh` if available)
- [ ] `node tests/fwlive-view-logging-toggle.test.js`
- [ ] `python3 tests/fwlive-jshn-compat.test.py`
- [ ] `./scripts/fwlive-test.sh` (or wait for CI)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved filter chip rendering so labels and negated filters display correctly.
  - Safely handles hostile filter, log, and hash values without injecting markup.
  - Ensures failed temporary-file creation, locking, or baseline snapshots fail safely without partial changes.
  - Clarified backend display labels and strengthened RPC method validation.

- **Tests**
  - Added coverage for chip rendering, hash restoration, shell safety, log output, protocol filters, ACL behavior, and backend labels.
  - Added compatibility checks for the address-resolution limit.

- **Documentation**
  - Updated the security review ledger and corrected a reference.


lucas-albers-lz4:
## Review — remaining #370 host coverage (`e1e0ee94`)

Independent pass over the diff, the shipped code it targets, and the acceptance criteria in #390–#393. Verdict: **the chips fix is correct and the new assertions discriminate, but one new failure injection does not reproduce under root, and two documentation statements now contradict the tree.** Both fixes are small and land in the same two files.

All touched suites were run on this branch; CI is green on this head.

### What I verified as correct (no action)

| Claim | Evidence |
|---|---|
| `chips.js` now keys the `contains` branch off `spec.key`, display off `spec.label` | `chips.js:19-40`; the only behaviour that changes is `q`, whose view label is `search` (`view/status/fwlive.js:164-173`). `src`/`dst`/`proto`/`sport`/`dport` labels equal their keys, so their rendering is byte-identical to before. |
| Single call site, no mirror to drift | `chipValueNodes` appears twice in the file (definition + the one call at `:74`); a repo-wide grep finds no second copy (`core/` holds only `fwlive-log.js`). No `gen-all.sh` involvement. |
| The new chip test discriminates | Red-proof: reverting the call to `chipValueNodes(spec.label, val)` fails the suite with `q must show contains, got: ≠search: not drop`. |
| The symlink-directory test targets the right guard | Red-proof: dropping `[ -L "$FILTER_TMP_DIR" ]` from `fwlive-log-filter.sh:49` fails the new test. The fixture is honest — the symlink target carries the sticky bit, so only the `-L` branch can produce `filter_tempfile_failed`. |
| Lock-failure test is uid-independent | Symlinked lock file → `{"error":"lock_failed"}` and no UCI mutation, reproduced as root and as an unprivileged user. |
| ACL arrays match the shipped grant | `read` = `logging_status, poll, resolve, rules`; `write` = `disable_wan_logging, enable_wan_logging` — equal to the ACL file and to the `list` parity assertion that is retained. The `ubus log.*` prelude invariant is untouched. |
| Table sink assertions are exact, not substring | `assertSinkOnlyTbodyClear` pins `deepStrictEqual(collectInnerHTMLWrites(host), [''])` and the recursion is centralised in the harness (`luci-e-harness.js:246`), so a nested bare-string child would be caught. |
| Harness consolidation | `fwlive-e-harness.test.js` and `modules-smoke` now share the harness helpers instead of carrying private copies; `proto-filter` uses the recursive form. |
| Suite results on this host | `chips-hash`, `proto-filter`, `modules-smoke`, `e-harness`, `view-logging-toggle`, `shell-filter` (`SH=sh` **and** `SH='busybox sh'`), `jshn-compat` (all five releases) pass. `fwlive-rpcd-security.test.js` cannot boot here, but it fails identically on `master` (dash `export -n` + jshn-variant precondition in its own prelude) — not a regression from this branch. |
| Issue scope | #393's five criteria are all present; #391 (filter mktemp, symlink, both toggle replies, no-mutation) and #392 R4a are met; #390 is met at host level. `Refs` only — nothing auto-closes, and the deferrals in the PR body match issues that stay open. |

### Findings

**1. `tests/fwlive-rpcd-security.test.js:647` — the baseline-snapshot failure injection does not reproduce under root (fix before merge).**
`chmod 0o555` on the baseline directory stops a normal user but not root, which bypasses DAC. Reproduced on this branch:

```
root  : AssertionError: 'firewall_reload_failed' !== 'baseline_snapshot_failed'   (test FAILS)
uid 65534: rpcd/fwlive: 398: cannot create .../baseline/wan-log-baseline: Permission denied
           driver: both toggle tests passed                                        (intended path)
```

Under root the snapshot **succeeds**, the toggle proceeds and later fails on the firewall reload, so the case never exercises `maybe_snapshot_wan_log_baseline` (`fwlive-logging.sh:391-401`) at all. CI runs as a non-root user, so it stays green there; a root or container run of `./scripts/fwlive-test.sh` gets a false red, and the AC "assert no partial staging … on failure" is then asserted against the wrong failure. `test-approach.md` asks for exactly this shape of injection — *"inject dependency failures through PATH, fixture, or seam controls"* — so the fix is to use a control no uid can bypass.
**Verified fix:** make the baseline path's parent a regular file, so `mkdir -p` fails for everyone:

```js
fs.writeFileSync(baselineDir, 'not a directory\n');   // instead of mkdirSync + chmod 0o555
const baselineFile = path.join(baselineDir, 'wan-log-baseline');
```

With that patch the test passes as root (`fix-check: baseline test passed as uid=0`) and still passes unprivileged. The `chmod` in the `finally` block then becomes unnecessary.

**2. `docs/developer/security-review.md:285` — the `chips.js` citation is stale, and the delta for this PR is missing (fix, two lines).**
Line 285's non-findings list cites `chips.js:94` as one of the four bare-identifier `E()` sites. This branch inserts two lines above that region, so the referenced text now sits at `:96` (I compared them: `master:94` and `head:96` are the same line, `);`); `head:94` is now the `['≠']` literal. A citation the next audit pass is told to trust should not point at the wrong construct.
The same file requires a delta: feature PRs touching the rpcd plugin, the ACL, or the shell helpers update it in the same PR (`:223`), and rule 5 requires correcting any claim this PR falsifies (`:216-221`). Worth noting while adding it: the existing `2026-09-19 #370 delta` (`:30`) states that host coverage "pins exact rpcd/ACL method parity" — on `master` `testAclMethodParity` only asserts union, disjointness and de-duplication, so that claim is only made true by this branch's `deepEqual` arrays. Recording it here closes the gap instead of leaving the ledger ahead of the code.

**3. `tests/fwlive-jshn-compat.test.py:65` — the new assertion pins source text, not behaviour (optional).**
`assert 'RESOLVE_MAX=32' in source` asserts that a literal string exists in the rpcd plugin. The behavioural cap is already pinned two lines below (33 addresses → 32 names, `192.0.2.33` absent), so this adds no detection: it breaks on a harmless refactor (computing the cap, renaming it, moving it behind a guard) while passing if the cap is applied at the wrong place. Either drop it and cite the constant in the comment above, or assert the degraded reply shape instead.

**4. `tests/fwlive-chips-hash.test.js:19` — the chip-field fixture mirrors the view without a parity check (nit).**
`CHIP_FIELDS` duplicates `FILTER_CHIP_FIELDS` (`view/status/fwlive.js:164-173`). The `q` → `search` label mismatch is precisely what the fix keys on, so if the view's labels change, this fixture silently weakens the test's premise instead of failing. The view is already loaded in this file for the `applyHash` cases — `loadFwliveView().view.FILTER_CHIP_FIELDS` removes the mirror.

**5. `tests/fwlive-chips-hash.test.js:118-133` — `withSelectOptions` patches a harness gap in test code (nit).**
`proto.js:30` reads `sel.options.length` / `sel.options[i].value`; the LuCI-accurate harness port does not model `options`, and `sel.options = opts` here synthesises it from the rendered `<option>` children. That belongs in `tests/lib/luci-e-harness.js` as an `options` accessor over `<option>` children, so the next select-driving test does not re-invent it.

**6. `tests/fwlive-modules-smoke.test.js:387-397` — the payload loop is unreachable (nit).**
Once `assert.deepStrictEqual(collectInnerHTMLWrites(host), [''])` passes, the following `for` loop cannot fail. Keep the array-shape assertion and drop the loop and the `payload` parameter.

**7. `tests/fwlive-chips-hash.test.js:73` — raw `_innerHTMLWrites` use next to the new helper (nit).**
`collectInnerHTMLWrites` is imported at `:16` and used at `:45`; the same assertion can be `collectInnerHTMLWrites(host).length >= 1`.

### Recommendation

Fold **1** (uid-independent injection) and **2** (citation + delta) before merge — one commit covers both, and 2 is the only part of the AC set that this branch currently contradicts. **3** is a judgement call on the greppability intent; **4–7** are cheap and fit the same pass if the file is open anyway. Nothing here blocks the `chips.js` fix itself.

Not reviewed: the deferred lanes (#389 installed lifecycle, #378 QEMU, #392 R4b session ACL, #390 mocked-browser fixtures) — item 5 of `test-approach.md` requires those to stay open with an owner, which the PR body already does.

lucas-albers-lz4:
Follow-up for the final review is pushed in `f04a0a4`.

Resolved:
- made the baseline-snapshot failure fixture deterministic for root by using a regular file at the directory path;
- made the BusyBox mktemp failure and symlink cleanup fixtures deterministic and safe;
- recorded the #394 security-review delta and corrected the `chips.js` citation.

The optional review cleanup items are also included in the PR branch. Validation passed with `./scripts/fwlive-test.sh`, including the focused root and BusyBox paths. The installed-lifecycle, installed-session, QEMU, and mocked-browser follow-ups remain deferred as noted in the review.

PR #401, opened 2026-09-20
test: add installed ACL session smoke for #392

lucas-albers-lz4:
## Summary

- add a reusable QEMU smoke for the installed LuCI/uhttpd `/ubus` ACL boundary
- authenticate one temporary user with `luci-app-fwlive` and one without it
- exercise read/write fwlive methods and assert the no-grant session is denied
- restore `/etc/config/rpcd` on exit; keep the host harness separate from live evidence

Refs #392

## Validation

- `bash tests/qemu-acl-session-harness.test.sh`
- `./scripts/fwlive-test.sh`
- `OPENWRT_SSH_PORT=2222 OWRT_HOSTFWD_HTTP=8080 ./scripts/qemu-acl-session-smoke.sh`

Live smoke passed on OpenWrt 24.10.8 x86_64; the dated evidence record is in the stacked follow-up PR.

## Summary by CodeRabbit

* **Tests**
  * Expanded QEMU ACL session smoke-test coverage to verify permitted and denied firewall-live operations, including `file.list` access and logging restrictions.
  * Improved test reliability with login retries, explicit JSON-RPC denial checks, source-state reporting, and stronger configuration backup and restoration validation.
  * Added automated checks confirming required ACL scenarios and test harness configuration.
  * Included the ACL session harness in the firewall live-test suite.


lucas-albers-lz4:
@coderabbitai rate limit

scripts/qemu-acl-session-smoke.sh
```diff
@@ -0,0 +1,196 @@
...
+trap cleanup_guest EXIT
+
+ssh_guest 'echo connected' >/dev/null 2>&1 \
+	|| die "SSH unreachable — start QEMU and install fwlive first"
+ssh_guest 'test -x /usr/libexec/rpcd/fwlive && test -x /usr/sbin/uhttpd' \
+	|| die "guest needs installed fwlive and uhttpd"
+
+BACKUP="$(ssh_guest 'mktemp /tmp/fwlive-acl-session.XXXXXX')"
+[[ -n "$BACKUP" ]] || die "could not allocate guest rpcd backup"
+ssh_guest "cp /etc/config/rpcd '$BACKUP'"
```
lucas-albers-lz4: **Bug:** `BACKUP` is assigned before `/etc/config/rpcd` is copied into it.

`cleanup_guest` treats a non-empty `BACKUP` as a valid restore source. If this script dies after `mktemp` and before `cp` (SSH blip, `set -e` on the copy, later `die` if you reorder), the EXIT trap does `cp '$BACKUP' /etc/config/rpcd` from an empty tempfile and restarts rpcd.

Assign `BACKUP` only after the copy is verified, e.g. `ssh_guest "cp /etc/config/rpcd '$tmp' && test -s '$tmp'"` then `BACKUP=$tmp`. In the trap, refuse to restore unless the backup file is non-empty.

scripts/qemu-acl-session-smoke.sh
```diff
@@ -0,0 +1,196 @@
...
+
+cleanup_guest() {
+	set +e
+	if [[ -n "$BACKUP" ]]; then
+		if [[ "$ORIGINAL_WAN_LOG" == true ]]; then
+			ssh_guest 'ubus call fwlive enable_wan_logging >/dev/null 2>&1' || true
+		elif [[ "$ORIGINAL_WAN_LOG" == false ]]; then
+			ssh_guest 'ubus call fwlive disable_wan_logging >/dev/null 2>&1' || true
+		fi
+		ssh_guest "cp '$BACKUP' /etc/config/rpcd && /etc/init.d/rpcd restart; rm -f '$BACKUP'" \
```
lucas-albers-lz4: Restore should `uci revert rpcd` before copying the backup back.

`uci add rpcd login` stages into `/tmp/.uci/rpcd`. If we exit before `uci commit`, the on-disk file is still original, but libuci (and therefore `rpcd restart`) will still see the overlay logins. `cp` alone does not drop that overlay.

Suggested order: `uci revert rpcd`, copy backup if it is non-empty, restart rpcd, remove the tempfile.

scripts/qemu-acl-session-smoke.sh
```diff
@@ -0,0 +1,196 @@
...
+}
+
+assert_denied() {
+	local label="$1"
+	local response="$2"
+	local code
+	code="$(rpc_status_code "$response")"
+	[[ "$code" != 0 ]] || die "$label should be denied, got: $response"
+	ok "$label denied (JSON-RPC result=$code)"
+}
```
lucas-albers-lz4, line 264: `[[ "$code" != 0 ]]` is too weak for an ACL oracle. #392 is specifically permission denial, not “any failure.”

uhttpd/ubus Access denied is JSON-RPC `-32002` (or ubus status `6` if the body uses `result[0]`). Treat only those as a pass. Object-not-found, method-not-found, parse errors, and HTTP/JSON wrapper failures should `die`.

The live 24.10.8 run already showed `-32002`; pin that so a broken guest cannot look like a successful deny.

scripts/qemu-acl-session-smoke.sh
```diff
@@ -0,0 +1,196 @@
...
+echo "acl-session smoke authenticated user=${GRANT_USER}"
+assert_allowed "${GRANT_USER}: fwlive.logging_status" "$(call_fwlive "$GRANT_SID" logging_status)"
+assert_allowed "${GRANT_USER}: fwlive.rules" "$(call_fwlive "$GRANT_SID" rules)"
+assert_allowed "${GRANT_USER}: fwlive.enable_wan_logging" "$(call_fwlive "$GRANT_SID" enable_wan_logging)"
+assert_allowed "${GRANT_USER}: fwlive.disable_wan_logging" "$(call_fwlive "$GRANT_SID" disable_wan_logging)"
+
+DENY_SID="$(login_session "$DENY_USER")"
+echo "acl-session smoke authenticated user=${DENY_USER}"
+assert_denied "${DENY_USER}: fwlive.logging_status" "$(call_fwlive "$DENY_SID" logging_status)"
+assert_denied "${DENY_USER}: fwlive.enable_wan_logging" "$(call_fwlive "$DENY_SID" enable_wan_logging)"
```
lucas-albers-lz4, line 285: This does not cover the #392 method list.

Grant session currently checks `logging_status`, `rules`, `enable_wan_logging`, `disable_wan_logging`. Missing:
- `poll` and `resolve` (empty `{}` is enough; both default-handle missing `addresses`)
- `log.read` must be **denied** for the grant session (security-model §6: sessions never receive `ubus log.read`)

Deny session only checks `logging_status` + `enable_wan_logging`. That is a representative pair, but the grant-side `log.read` check is the one host JSON tests cannot replace.

tests/qemu-acl-session-harness.test.sh
```diff
@@ -0,0 +1,27 @@
...
+bash -n "$SCRIPT"
+shellcheck --severity=warning "$SCRIPT"
+"$SCRIPT" --help >/dev/null
+
+grep -Fq 'http://${HOST}:${HTTP_PORT}/ubus' "$SCRIPT" \
+	|| { echo "helper must exercise the HTTP ubus endpoint" >&2; exit 1; }
+grep -Fq 'luci-app-fwlive' "$SCRIPT" \
+	|| { echo "helper must assign the shipped ACL group" >&2; exit 1; }
+grep -Fq 'assert_denied' "$SCRIPT" \
+	|| { echo "helper must assert the no-grant session" >&2; exit 1; }
```
lucas-albers-lz4, line 23: These greps will not fail if `poll`, `resolve`, or `log.read` stay omitted from the live smoke. If the helper is the R4b oracle, the host harness should at least require those method strings (and a `log.read` deny) so the coverage cannot regress without `fwlive-test.sh` noticing.

Keep the existing disclaimer: this still must not be treated as live ACL enforcement.

lucas-albers-lz4 (commented):
## Review

The split is right: host harness stays syntax/boundary-only, and the live proof goes through LuCI/uhttpd `/ubus` rather than root `ubus call`. A few holes keep this from matching the #392 R4b contract.

### Blocker
- **`cleanup_guest` can wipe `/etc/config/rpcd`.** `BACKUP` is set as soon as `mktemp` returns, before the copy from `/etc/config/rpcd` succeeds. On any exit between those lines the trap copies an empty tempfile over rpcd and restarts it.

### Should fix before treating this as R4b proof
- **`assert_denied` accepts any non-zero JSON-RPC code.** A missing `fwlive` object, method-not-found, or wrapper error would pass. Pin Access denied (`-32002` / ubus `6`).
- **Grant path does not exercise the full ACL surface.** #392 asks for read `rules`, `poll`, `resolve`, `logging_status`; write `enable_wan_logging`, `disable_wan_logging`; and the session must not get `log.read`. This smoke allows four methods and denies two. `poll`/`resolve` with `{}` are enough to prove ACL allow; add a grant-session `log.read` deny.
- **Cleanup should `uci revert rpcd` before restore.** If the script dies after `uci add` and before `uci commit`, `/tmp/.uci/rpcd` still has the test logins. `cp` of the original file plus `rpcd restart` still loads the overlay.

### Nits
- The deny user has no ACL groups at all. That proves “empty session cannot call fwlive”, not “a normal LuCI login without `luci-app-fwlive` cannot.” Granting `luci-base` (or the stock read set minus `luci-app-fwlive`) would match the issue wording more closely.
- Host grep checks do not pin `poll`, `resolve`, or `log.read`, so the coverage gaps above would not fail `fwlive-test.sh`.

CI host tests are green; this is about the live oracle, not syntax.

lucas-albers-lz4:
@coderabbitai rate limit

lucas-albers-lz4:
Addressed the review findings in commit `310d6c2686`:

- arm the restore trap only after a successful rpcd backup copy
- `uci revert rpcd` before restoring the file
- pin HTTP Access denied to JSON-RPC `-32002`
- cover all six fwlive methods and deny `log.read`
- give the negative session `luci-base` without `luci-app-fwlive`
- extend the host harness pins

Validated with the live OpenWrt 24.10.8 QEMU session smoke and the full `./scripts/fwlive-test.sh` run. The stacked #402 artifact was refreshed to match the new evidence.

lucas-albers-lz4:
@coderabbitai rate limit

lucas-albers-lz4:
@coderabbitai rate limit

lucas-albers-lz4:
@coderabbitai rate limit

lucas-albers-lz4 (commented):
## Review

Verified at `6fabcd91` (branch head, unchanged since the review above): the shipped ACL sets (`openwrt-feed/luci-app-fwlive/root/usr/share/rpcd/acl.d/luci-app-fwlive.json` → read `rules, poll, resolve, logging_status`; write `enable_wan_logging, disable_wan_logging`), the smoke's actual method coverage, `bash -n` + `shellcheck --severity=warning` (exit 0), and the trap/restore semantics line by line. The housekeeping claim holds: the host test and the installed evidence are cleanly separated.

The split is still the right call. What blocks this from being R4b proof is that the deny assertion cannot tell an ACL denial from any other failure, and one cleanup path can destroy the guest's rpcd config.

### Blocker

- **`cleanup_guest` can copy an empty file over `/etc/config/rpcd`.** `BACKUP` is assigned from `mktemp` at `:100`, and the copy from `/etc/config/rpcd` is the *next* statement at `:102`. If `:102` fails (SSH drop, transient error) the shell exits under `set -e` and the `EXIT` trap at `:93` runs `cleanup_guest`, whose `cp '$BACKUP' /etc/config/rpcd` at `:88` puts a zero-byte file in place of rpcd's config and restarts the service. The `[[ -n "$BACKUP" ]]` guard at `:82` checks that the *path string* is non-empty — it does not prove the copy landed, so it does not close this window. Narrow window, destructive outcome. Fix: only materialise the sentinel after a successful copy (`tmp="$(ssh_guest mktemp /tmp/fwlive-acl-session.XXXXXX)"; ssh_guest "cp /etc/config/rpcd '$tmp'" && BACKUP="$tmp"`), or gate the trap on a separate `BACKUP_READY` flag set post-copy.

- **`assert_denied` accepts any non-zero code — it is not an ACL oracle.** `:171-178` tests only `code != 0`, and `rpc_status_code` (`:158-160`) resolves `.result[0] // .error.code // 255`. So a method-not-found, a missing `fwlive` object, a wrong-argument error or a malformed reply all satisfy "denied" — the assertion cannot distinguish an access-control denial from an unrelated dispatch failure. (Transport failures do abort correctly via `curl -fsS` + `set -e`; the false-pass is specifically for valid JSON-RPC error responses.) The artifact records `-32002`, so pin exactly that (ubus `6` also acceptable), require an error envelope rather than a `.result[0]` code, and treat the `255` fallback as a harness error instead of a denial.

### Should fix before this counts as R4b

- **Method coverage does not match the #392 contract.** `:186-189` exercises `logging_status`, `rules`, `enable_wan_logging`, `disable_wan_logging`; `:193-194` checks denial for `logging_status` and `enable_wan_logging` only.

	| ACL method | Scope | Grant session | Deny session |
	|---|---|---|---|
	| `rules` | read | asserted allowed | not exercised |
	| `poll` | read | **not exercised** | not exercised |
	| `resolve` | read | **not exercised** | not exercised |
	| `logging_status` | read | asserted allowed | asserted denied |
	| `enable_wan_logging` | write | asserted allowed | asserted denied |
	| `disable_wan_logging` | write | asserted allowed | not exercised |

	`poll`/`resolve` are also the two methods that carry the interesting resolution paths. The `log.read` invariant is still host-only, as the artifact states — but that means it is not covered at the installed boundary.

- **The deny user has no ACL groups at all.** `add_login` (`:112-124`) adds `read`/`write` lists only for `$GRANT_USER`, so the no-grant session proves "an empty-ACL login cannot call fwlive", not "a LuCI session without `luci-app-fwlive` cannot". Grant the deny user the stock read set minus the fwlive group to match the issue wording.

- **Cleanup never clears the UCI delta.** `uci add rpcd login` (`:115`) stages the test logins in the guest's `/tmp/.uci/rpcd`; if the script dies between `:115` and `uci commit rpcd` (`:128`), the trap's `cp` restore does not touch that overlay, so the stale logins can be flushed later by any commit and are inherited by a re-run. Add `ssh_guest 'uci -q revert rpcd 2>/dev/null' || true` to `cleanup_guest`.

- **The restore is unverified, and its failure destroys the backup.** `cleanup_guest` sets `set +e`, discards all output and never inspects the result, so a run whose restore failed still exits `0` — which is what `security-review.md` and the artifact assert did not happen. Worse, `:88` chains `cp '$BACKUP' /etc/config/rpcd && /etc/init.d/rpcd restart; rm -f '$BACKUP'`: because the separator before `rm` is `;`, a failed copy or restart still deletes the only backup and skips the restart, silently. Verify the restore (e.g. compare a hash of the restored file) and only `rm` on success.

- **The host harness pins no method or denial code.** `tests/qemu-acl-session-harness.test.sh:18-25` asserts four literal tokens (the `/ubus` URL, `luci-app-fwlive`, `assert_denied`, `cleanup_guest`). Dropping `poll`/`resolve`/`log.read`, or weakening `assert_denied` back to "non-zero", would still pass `./scripts/fwlive-test.sh`. Pin the method names and the expected denial code so the coverage above cannot regress silently.

### Nits

- `sleep 1` after the rpcd restart (`:129`) is a fixed wait; a short retry against `/ubus` would be steadier on a cold guest.
- `SOURCE_SHA` (`:16`) is the host `HEAD`, and the banner that prints it (`:180`) comes after the assertions — an early failure loses the source binding, and a dirty worktree records a commit that does not describe the deployed bytes.
- `shellcheck` flags SC2029 (`:77`) and SC2016 (`:180`) at info level; the harness runs `--severity=warning`, so the gate passes. Consistent with other lab scripts, but tidier if escaped.
- The default `TEST_PASSWORD` at `:20` is a static lab string in a public repo. Safe under the `[A-Za-z0-9._-]` allowlist, but the artifact's description of it is wrong (see #402).

### Not findings (checked, so you don't have to re-litigate them)

- `StrictHostKeyChecking=no` / `UserKnownHostsFile=/dev/null` in `SSH_OPTS` is consistent with ~20 sibling `scripts/qemu-*.sh` lab harnesses, and the ledger's opt-in control binds `scripts/agent-build-and-deploy.sh` (which moves artifacts). This script transfers none, so nothing is violated.
- The `EXIT`-only trap does **not** skip the restore on Ctrl-C: bash runs an `EXIT` trap on SIGINT and SIGTERM (verified). Only SIGKILL escapes.
- The `jsonfilter` dependency dying before the overlay check is not a restore hazard — the copy at `:102` has already succeeded by then, so the trap has a valid backup.

### Confirmed good

- HTTP `/ubus` (not root `ubus call`) is the correct session boundary: `:131-136`.
- The host/installed split is explicit and the harness disclaims live enforcement: `tests/qemu-acl-session-harness.test.sh:1-7`.
- WAN-log snapshot/restore is correct on the happy path: `:103-107`, `:83-87`.

lucas-albers-lz4:
@coderabbitai rate limit

lucas-albers-lz4:
@coderabbitai rate limit

lucas-albers-lz4:
@coderabbitai rate limit

lucas-albers-lz4:
@coderabbitai rate limit

lucas-albers-lz4:
@coderabbitai rate limit

lucas-albers-lz4:
@coderabbitai rate limit

lucas-albers-lz4:
@coderabbitai review

lucas-albers-lz4:
Addressed the additional cleanup-integrity review in commit `7702ff3ca0`:

- backup trap arms only after a non-empty copy and validated SHA-256
- cleanup reverts staged UCI state, restores/restarts rpcd, verifies the restored hash, retains the backup on failure, and turns a cleanup failure into a non-zero run
- WAN logging restore failures are no longer ignored
- fixed sleep replaced with ten login/readiness retries
- `assert_denied` now reads only the JSON-RPC error envelope and pins `-32002`
- default lab passwords are per-run; the source banner is emitted early and records `source_state=clean|dirty`
- shellcheck informational notes are documented/suppressed where remote expansion is intentional

Validated with the clean-source QEMU smoke on OpenWrt 24.10.8: all six fwlive allows, `log.read` denial, ordinary `luci-base` denial, and hash-verified cleanup. Full `./scripts/fwlive-test.sh` also passes. #402 now records source `7702ff3ca0` and the verified cleanup semantics.

lucas-albers-lz4:
@coderabbitai rate limit

lucas-albers-lz4 (commented):
## Review

Verified at `7702ff3c` (branch head) against the shipped ACL, not just the diff. The folds close the previous round's blockers:

- `cleanup_guest` arms the sentinel only after `cp` + `test -s` + a 64-hex hash (`scripts/qemu-acl-session-smoke.sh:183-200`) — the empty-file-over-rpcd window is gone.
- `assert_denied` reads `.error.code` and pins `-32002` (`:288-295`); the `255` fallback is now a harness error rather than a denial.
- Coverage matches the shipped ACL (`root/usr/share/rpcd/acl.d/luci-app-fwlive.json`: read `rules, poll, resolve, logging_status`; write `enable_wan_logging, disable_wan_logging`) — six grant calls, a grant-session `log.read` deny, and two deny samples.
- Cleanup reverts staged UCI, restores, hash-verifies, and only then removes the backup (`:136-174`).
- `tests/qemu-acl-session-harness.test.sh` pins the six methods, the `log.read` call, and `-32002`, and is wired at `scripts/fwlive-test.sh:166` → CI (`.github/workflows/fwlive-test.yml:34`).

Local: `bash -n` clean, `shellcheck --severity=warning` exit 0, harness passes.

### Should fix before this counts as R4b

- **The deny lane has no positive control.** `:224` grants the deny login `read=luci-base`, but nothing exercises that grant — `:312-315` only assert denials. If `uci add_list` silently misses, or `luci-base` does not resolve in the guest, the deny user is an empty-ACL session and both assertions still pass: the same ambiguity the previous round flagged ("no ACL groups at all"), in weaker form. Add one allowed call for the deny session before the fwlive denials — `luci-base` reads `ubus file.list` and `ubus uci.get` (`modules/luci-base/root/usr/share/rpcd/acl.d/luci-base.json`), so e.g. `call_object "$DENY_SID" file list` with `{"path":"/etc/config"}` works as the control. Then "denied" means "authenticated LuCI user minus this ACL", which is what the doc sentences claim.

### Minor

- **`docs/developer/security-review.md` (#392 base line) overstates the deny subject:** "asserts the exact JSON-RPC Access denied code for an ordinary LuCI session without the app ACL". A stock LuCI login carries several ACL groups; this helper grants exactly one (`luci-base`), and the script comment at `:222-223` says "Match an ordinary authenticated LuCI user". #402's R4b row already states it correctly ("`luci-base` only") — align these two.

### Nits

- `uci revert rpcd` (`:144`) discards *any* staged rpcd edits, not only the helper's. Correct for cleanup, but one line in `usage()` would cover it.
- The harness greps tokens, not behaviour: an `assert_denied` weakened to accept any non-zero code would still pass all 35 lines as long as the literal `-32002` survives in the fail message. Low stakes for a host tripwire; if you want the lock to bite, assert the comparison shape or drive `assert_denied` with a fake `curl`.

### Confirmed good

- The host/installed split is explicit and the harness disclaims live enforcement (`tests/qemu-acl-session-harness.test.sh:1-7`).
- HTTP `/ubus` rather than root `ubus call`, and the grant-session `log.read` deny closes the one accepted control that was host-only until now.

PR #403, opened 2026-09-20
docs: align supported firewall contract

lucas-albers-lz4:
## Summary

- align README, requirements, and logging references with the supported 23.05+ firewall4/nft contract
- keep 21.02/22.03 material as historical notes without dead SDK/validation commands
- document the ACL-only menu prerequisite and the current `using fw4` / `admin/network/firewall/rules` UI contract
- replace stale ubus-only comments and record the registered-table diagnostic false-positive boundary
- make the current Unreleased support floor explicit without rewriting historical release entries

Refs #378

## Validation

- `./tests/fwlive-logging.test.sh`
- `./scripts/fwlive-linkcheck.sh` (429 internal, 71 external; 0 failures)
- `git diff --check`

No shipped behavior was changed; the shell change is comment-only.

PR #404, opened 2026-09-20
test: cover legacy table probe edge cases

lucas-albers-lz4:
## Summary

- add whitespace-only and blank-first-line/populated-second-line procfs fixtures for `legacy_iptables_active`
- preserve and assert unchanged readiness for diagnostic-only warnings
- record that registered table names are inventory evidence, not proof of active rules, so an empty built-in table can be a deliberate diagnostic false positive

Depends on #403
Refs #378

## Validation

- `./tests/fwlive-logging.test.sh`
- `git diff --check`

This remains host-only fixture coverage; no runtime behavior change.


lucas-albers-lz4 (commented):
## Review

Whitespace-only → no warning is the right oracle for `grep -q '[^[:space:]]'`, and it is isolated (IPv6 is still empty). Readiness pins are consistent with the existing diagnostic-only contract. CI `test` is green.

The blank-first-line case does not actually prove the grep-whole-file behavior it is meant to lock.

### Should fix
- After the IPv6-populated case, `LEGACY_IPV6` still contains `filter`. The next fixture only rewrites IPv4 to `\nfilter\n`. `assert_legacy_warning_state … present` would pass even if the probe read only the first IPv4 line (empty) and ignored the rest, because IPv6 already warns. Clear IPv6 before that case (`: >"$LEGACY_IPV6"`), the same way the IPv6-only case clears IPv4.

### Ledger nit
- **Method** lists the new fixtures accurately. **Result** then describes a different residual (named empty built-in table / no active rules). That residual is already the `printf 'filter\n'` case; these fixtures are about whitespace and later-line names. Keep Result on the parse edges, or say the empty-table FP is unchanged and still not a rules-vs-names test.

tests/fwlive-logging.test.sh
```diff
@@ -118,6 +122,10 @@ printf 'filter\n' >"$LEGACY_IPV6"
 out=$(build_logging_status_json)
 assert_legacy_warning_state "$out" present "IPv6 legacy table"
 assert_ready_unchanged "$out" "IPv6 legacy table"
+printf '\nfilter\n' >"$LEGACY_IPV4"
```
lucas-albers-lz4, line 126: This case is not isolated.

The previous block left `filter` in `$LEGACY_IPV6`. `legacy_iptables_active` ORs both files, so a first-line-only `read` implementation (the sketch in #378) would still warn from IPv6 and this assertion would still pass.

Clear IPv6 first, matching the IPv6-only setup:

```sh
: >"$LEGACY_IPV6"
printf '\nfilter\n' >"$LEGACY_IPV4"
```

That is the fixture that distinguishes whole-file `grep -q '[^[:space:]]'` from `IFS= read -r` on line 1.

docs/developer/security-review.md
```diff
@@ -581,14 +581,18 @@ diagnostic-only: it does not affect `ready`, `blockers`, or the enable gate.
...
-change was made.
+**Method.** Host fixtures cover missing, empty, whitespace-only,
+IPv4-populated, blank-first-line/populated-second-line, IPv6-populated, and
+both-populated table-name files through `FWLIVE_IP_TABLES_NAMES_PATH` and
+`FWLIVE_IP6_TABLES_NAMES_PATH`. Empty fixture files remain exported so later
+`logging_status` tests do not read host procfs; no host kernel state or
+privilege is required.
+
+**Result.** Focused logging tests pass. The warning intentionally reports any
+registered table-name entry, not active rules; a loaded legacy module with an
```
lucas-albers-lz4, line 593: These new fixtures do not show “a loaded legacy module with an empty built-in table.” They show:
- whitespace-only file → no warning
- blank line then a table name → warning (once IPv6 is cleared)

An empty `filter` table is indistinguishable here from a populated one; both are `filter\n` in `ip_tables_names`. That inventory-vs-rules residual was already covered by the IPv4-populated case.

Point Result at the parse edges, and keep the empty-table FP as a one-line accepted residual if you still want it on the record.

lucas-albers-lz4 (commented):
## Review

Verified at `14298f87`. The new coverage is real — I fault-injected the probe instead of trusting the green suite:

| Injection into `legacy_iptables_active` (`fwlive-logging.sh:583-592`) | Result at this head |
|---|---|
| `[ -s "$_path" ]` instead of `grep -q '[^[:space:]]'` | **fails**: `whitespace-only legacy table file: must not warn` |
| `head -n1 "$_path" \| grep -q '[^[:space:]]'` (first line only) | **fails**: `blank-first-line legacy table: missing from warnings` |
| the same first-line-only probe against `origin/master` | passes (exit 0) — the fault was invisible before this PR |

The tree was restored byte-identical after each injection (`git diff --stat` empty), and the suite passes at head with `sh` → bash.

The previous round's isolation gap is closed: `: >"$LEGACY_IPV6"` before the blank-first-line case (test file ~`:125-127`) leaves IPv4 `\nfilter\n` as the only possible warning source, so the assertion now genuinely pins the grep-whole-file behaviour it claims. The second injection is what proves it — and the master row above shows the pre-PR suite passed that same fault.

The ledger rewrite is accurate: the probe warns on any non-whitespace byte in `/proc/net/{ip,ip6}_tables_names`, so a registered table is inventory rather than proof of active rules, and readiness is unchanged in every fixture (`assert_ready_unchanged` on each case).

### Nit

- Fixtures cover whitespace-only and blank-first-line, but not a CR-only file, which `[^[:space:]]` also treats as "no warning". Optional — noted only because the probe's whole contract is now fixture-asserted.

### Environment note (not a finding)

Same as #403: `./tests/fwlive-logging.test.sh` fails on this box because `/bin/sh` is dash (`export: Illegal option -n` → `rpcd __selftest`), identically on `origin/master`; CI is unaffected.

### Confirmed good

- Whitespace-only is the right oracle for a `[^[:space:]]` grep — the fixture is non-empty (6 bytes), so it discriminates `-s` from a content check.
- Fixture files stay exported and empty after the block, keeping later `logging_status` pins host-independent.

Issue #389, opened 2026-09-19
validation: record installed lifecycle and 23.05 IPK evidence


lucas-albers-lz4:
Parent review: #370 (R8/R9a)

Capture the remaining installed and package-artifact evidence required before the upstream cut.

Acceptance criteria:
- record 24.10 armsr installation, read methods, write behavior, and uninstall/baseline restoration;
- record 25.12 x86_64 APK installation, read methods, write behavior, apk del, and baseline restoration;
- inspect the 23.05 IPK payload, or explicitly document and approve representative IPK coverage;
- store dated artifacts and link them from the acceptance record.

## Residual: packaged prerm upgrade vs remove

Same-release reinstall must not undo WAN logging. That is the claim in `docs/user/installation.md` (“WAN zone `log` stays on across upgrades until you turn it off or uninstall”). It is not a 24.10→25.12 sysupgrade cell: users cannot `apk add` a 25.12 package onto a running 24.10 opkg system.

The contract lives in the packaged `prerm` argv/hook, not in `restore_wan_log_baseline()` itself. Host tests of the restore function and the R8 uninstall smokes (`opkg remove` / `apk del`) do not prove in-place upgrade. #398 records those uninstall cells and currently invokes restore on any live-root `prerm`/`pre-deinstall` (the old `$1 = remove` guard was dropped so apk uninstall works). That makes upgrade-in-place the remaining discriminating case.

- Host: run the packaged hook with `remove`, `upgrade`, and empty argv against a stubbed baseline. Discriminate: uninstall restores; upgrade and empty-argv-as-upgrade must not restore. If apk’s real `pre-deinstall` cannot tell upgrade from remove, record that as the product decision and pin the installed check below as the oracle.
- Installed, on the existing 25.12 x86_64 apk cell (and 24.10 opkg if the host oracle is not enough): enable logging, reinstall the same package in place (`apk add` / `opkg install` over the installed copy), assert the WAN log bit and `/etc/fwlive/wan-log-baseline` stay put. Then run the already-required uninstall/baseline restoration.

Fold into this issue’s artifacts. No new matrix cell.

lucas-albers-lz4:
Implemented and pushed the installed-package evidence fix in PR #398.

[...]

PR #405, opened 2026-09-20
fix: preserve WAN baseline across package upgrades

lucas-albers-lz4:
Refs #389

## Summary
- make the packaged prerm understand OpenWrt's generated opkg wrapper arguments and APK 3.0 version-valued pre-deinstall arguments
- skip upgrades, empty/unknown actions, staging roots, and preserve best-effort uninstall restoration
- test the generated IPK wrapper/body boundary and keep QEMU uninstall smokes on artifact maintainer scripts

## Verification
- `./scripts/fwlive-test.sh`
- `./scripts/docker-sdk.sh build --target x86-64 --version 24.10`
- `./scripts/docker-sdk.sh build --target x86-64 --version 25.12`
- QEMU 24.10.8: same-version `opkg install` preserved WAN logging; `opkg remove` restored it
- QEMU 25.12.5: same-version `apk add --force-reinstall` preserved WAN logging; `apk del` restored it

Final human review can run asynchronously; the implementation is not waiting on that review to continue the evidence stack.

## Summary by CodeRabbit

- **Bug Fixes**
  - Package removal hooks now restore the WAN log baseline only for confirmed uninstall actions.
  - Upgrade, reinstall, invalid, empty, and unsupported actions no longer trigger restoration.
  - Non-root staging installs safely skip runtime restoration steps.
  - Restoration failures remain best-effort and do not block package removal.

- **Tests**
  - Added lifecycle coverage for packaged IPK and APK uninstall behavior, including action handling and failure scenarios.
  - Lifecycle checks now run automatically across supported releases.


lucas-albers-lz4:
@coderabbitai review

tests/fwlive-package-lifecycle.test.sh
```diff
@@ -0,0 +1,183 @@
...
+			else
+				echo 'fwlive-package-lifecycle test FAIL: could not extract IPK control archive' >&2
+				exit 1
+			fi
+			tar -tzf "$CONTROL" | grep -qx './prerm'
+			tar -tzf "$CONTROL" | grep -qx './prerm-pkg'
+			tar -xzOf "$CONTROL" ./prerm >"$WORK/ipk-prerm"
+			tar -xzOf "$CONTROL" ./prerm-pkg >"$WORK/ipk-prerm-pkg"
+			grep -Fq 'default_prerm $0 $@' "$WORK/ipk-prerm"
+			grep -Fq 'restore_wan_log_baseline' "$WORK/ipk-prerm-pkg"
```
lucas-albers-lz4: **CONFIRMED / medium (Composer + Grok):** this IPK “wrapper/body boundary” check only greps for `default_prerm $0 $@` and `restore_wan_log_baseline`. That would also pass the old always-restore `prerm`. The action matrix below extracts the Makefile define, so a stale `out/` IPK or a packaging transform that dropped the new `case` still goes green.

Fix: run the extracted `./prerm-pkg` (and/or the generated wrapper) through the same upgrade/remove/empty/unknown/staging cases, and fail if the packaged body diverges from the Makefile hook.

scripts/fwlive-test.sh
```diff
@@ -191,6 +191,9 @@ bash tests/feed-publish-release-assets.test.sh
 echo "== fwlive package payload layout (R9a) ==" >&2
 bash tests/fwlive-package-payload.test.sh
 
+echo "== fwlive package lifecycle hook (#389) ==" >&2
```
lucas-albers-lz4, line 194: **CONFIRMED / medium (Grok):** default `./scripts/fwlive-test.sh` does not set `FWLIVE_REQUIRE_PACKAGE=1`, so CI skips artifact inspection whenever `out/` has no current IPK. The per-PR job that **does** build 24.10 IPK / 25.12 APK still only runs `fwlive-package-payload.test.sh`, not this lifecycle test. The wrapper/body check is not a gate.

openwrt-feed/luci-app-fwlive/Makefile
```diff
@@ -28,11 +28,30 @@ case "$${IPKG_INSTROOT}" in
...
+# A same-version APK reinstall uses post-upgrade and does not run this hook.
+[ "$${PKG_UPGRADE:-}" = 1 ] && exit 0
+case "$${2-}" in
+	remove) ;;
+	upgrade) exit 0 ;;
+	'')
+		case "$${1-}" in
+			remove) ;;
+			upgrade|''|[!0-9]*) exit 0 ;;
+			[0-9]*.[0-9]*) ;;
```
lucas-albers-lz4: **QUESTION / low (Composer + Grok):** `[0-9]*.[0-9]*` is a glob, so `.` matches any character. Current `0.1.44-r1` still hits this arm; `123` correctly fail-closes via `*)`. Host tests never set `PKG_UPGRADE=1` or the real opkg argv `upgrade <new-version>` — both still fail closed today, but they are unproven.

lucas-albers-lz4 (commented):
## Composer + Grok + Bugbot review (pre-fix)

Hook logic looks sound for opkg wrapper args and APK 3.0 version-valued `pre-deinstall`: upgrades skip restore, remove still restores, staging/`IPKG_INSTROOT` fail closed. Dropping the planted always-restore `prerm` in `qemu-install-fwlive.sh` strengthens uninstall smokes. Bugbot found no bugs. No ACL / DOM-sink / `log.read` issues.

The remaining work is proof, not the `case` dispatch:

| Severity | Location | Finding | Label |
| --- | --- | --- | --- |
| medium | `tests/fwlive-package-lifecycle.test.sh` | Action matrix runs a Makefile-extracted hook, not packaged `prerm-pkg`. IPK checks `grep` for `default_prerm $0 $@` and `restore_wan_log_baseline`, which would also pass the **old always-restore** body. | CONFIRMED |
| medium | `scripts/fwlive-test.sh` / CI payload job | Default host tests skip artifact inspection without `FWLIVE_REQUIRE_PACKAGE=1`. The job that builds real IPK/APK still runs payload tests, not this lifecycle test. | CONFIRMED |
| low | Makefile / host test | `PKG_UPGRADE=1` and real `prerm upgrade <new-version>` are not exercised; APK glob `[0-9]*.[0-9]*` is looser than a dotted-version regex. | QUESTION |

Fix after the next review: execute packaged `prerm-pkg` in the host matrix (not just grep), and gate artifact inspection on the package-build job.

lucas-albers-lz4 (commented):
## Review — mechanism verified end-to-end; the packaged-artifact oracle cannot fail

I re-derived the argument contract from the real OpenWrt code on disk instead of trusting the comments:

- opkg calls the script **with the action**: `pkg_run_script(pkg, "prerm", "remove")` (`libopkg/opkg_remove.c:301`).
- the installed wrapper is `. ${IPKG_INSTROOT}/lib/functions.sh; default_prerm $0 $@` (verbatim from a real 24.10 `/usr/lib/opkg/info/base-files.prerm`), and the real `default_prerm` sources `<pkg>.prerm-pkg` in a subshell.
- instrumenting that chain, the body receives **`$1=/usr/lib/opkg/info/luci-app-fwlive.prerm`** and **`$2=remove`**.

Running the three real bodies (Makefile-extracted with the same `awk` as the test) through that real wrapper + `default_prerm`, with `IPKG_INSTROOT` empty exactly as on a router:

| body | `opkg remove` | `opkg upgrade` | remove + `PKG_UPGRADE=1` | no action arg | APK `pre-deinstall` (version as `$1`) |
|---|---|---|---|---|---|
| tag `v0.1.44` | **no** | no | no | no | no |
| `origin/master` (pre-fix) | YES | **YES** | YES | YES | YES |
| **this PR** | **YES** | **no** | **no** | no (fail-closed) | **YES** |

So the dispatch is correct for every form opkg can present, the `PKG_UPGRADE` arm is belt-and-braces (the action arg already covers upgrades), the staging guard behaves as documented, and the APK version-valued form works. `bash tests/fwlive-package-lifecycle.test.sh` passes here; CI `test` is green.

### Should fix — the ledger/hook story about the released package isn't accurate

At tag `v0.1.44` the body is `[ "$$1" = "remove" ] || exit 0`, and `$1` is the wrapper **path** — so through the installed wrapper that guard can never fire (row 1). Two possibilities, and both mean the released package did not restore on uninstall: the hook was absent from the artifact (that tag's `define` sat *after* `include $(TOPDIR)/feeds/luci/luci.mk`, which #398 moved above it), or it was present and never matched. Worth one command to settle it — inspect the published feed's `0.1.44-r1` IPK control archive for `./prerm` + `./prerm-pkg`.

Either way the record should say it, because the pre-#405 "#389 live opkg `remove` … verified to restore the WAN-log baseline" observation was taken against the lab-planted `prerm` that **this PR deletes** from `scripts/qemu-install-fwlive.sh`. As written, the ledger reads as though the released artifact had been proven.

Related, one sentence of residual: installs carrying the `master`-era always-restore body (lab builds from the first #389 attempt) run *that* body with `upgrade` on the next version bump and lose the baseline once — not preventable from inside the new package. Tagged releases carry the `v0.1.44` body, so field users are unaffected. Also note `docs/release.md:36` keeps `PKG_RELEASE:=1` and bumps the third octet at release time, so the fix reaches users only at the next `v*` bump.

### Should fix — the packaged-artifact check is non-discriminating, and CI never runs it (reproduced)

Both verified here:

- **It passes the body this PR replaces.** I built two `ar` IPKs — one with the pre-fix always-restore body, one with this PR's body — and `FWLIVE_REQUIRE_PACKAGE=1 FWLIVE_PACKAGE=… bash tests/fwlive-package-lifecycle.test.sh` printed `artifact OK` and exited 0 for **both**. `grep -Fq 'restore_wan_log_baseline'` is satisfied by the replaced body too.
- **Nothing in CI reaches that branch.** `.github/workflows/fwlive-test.yml:100-104` exports `FWLIVE_REQUIRE_PACKAGE=1 FWLIVE_PACKAGE=…` but runs only `tests/fwlive-package-payload.test.sh`; `scripts/fwlive-test.sh` runs the lifecycle test without the package env, so with no `out/` it silently skips — confirmed: with `FWLIVE_REQUIRE_PACKAGE=1` here it fails closed with `no current package artifact found`.

Fix: run the lifecycle test in the payload job with the same `FWLIVE_PACKAGE`/`FWLIVE_PACKAGE_FORMAT` env, and make the packaged check behavioural — run the packaged `prerm-pkg` through the action matrix, or assert the absence of the always-restore form — instead of grepping a string the old body also contains.

### Nit — the one hop the matrix never takes

`run_hook` calls the extracted body directly and `run_opkg_hook` passes `$1=path $2=action`, which matches reality (verified above), but no case runs the body *through* the wrapper + `default_prerm` — the hop that turns `$1` into the wrapper path and that broke `v0.1.44`. A ~10-line fixture (wrapper text at `/usr/lib/opkg/info/luci-app-fwlive.prerm` + the real `default_prerm`, `IPKG_INSTROOT` empty) closes that hole; happy to hand over the harness I used.

### Question — APK version-changing upgrade is indistinguishable

A version-shaped `$1` with empty `$2` is both the APK removal input (row 5) and what the hook would see if apk runs the old package's `pre-deinstall` on a **version-changing** upgrade. The APK cell here is same-version `apk add --force-reinstall`, and the 24.10 x86 IPK cell is an "up-to-date" no-op — so nothing covers the case the PR title names. Either add a lab cell that upgrades `0.1.44-r1` to a bumped release, or state in the evidence and the hook comment that only same-version reinstall was verified for APK.

### Confirmed good

- Deleting the planted `prerm` from `scripts/qemu-install-fwlive.sh` is the single most valuable change in this PR: it stops uninstall smokes from validating the source tree instead of the artifact.
- `tests/fwlive-package-lifecycle.test.sh` is a genuinely useful host gate — the action matrix, staging case, and best-effort-failure case all pin real behaviour.

lucas-albers-lz4 (commented):
## Verified findings — #405

Every item below was re-checked by hand against `78b984c5b4` (file/line, `grep`, and CI-config level). A full three-reviewer pass is queued behind the Go weekly quota reset and will follow; it re-fetches the branches first, so it will review whatever is pushed in the meantime.

The hook's argv contract itself is right, and I could not break it: both real shapes resolve correctly — opkg wrapper + action via `$2`, apk version-valued `$1` with empty `$2` — and `upgrade`, empty, non-version and unknown actions all fail closed (`openwrt-feed/luci-app-fwlive/Makefile:32-52`). Dropping the planted always-restore `prerm` from `scripts/qemu-install-fwlive.sh` is the right call: it was making the uninstall smokes validate the source tree instead of the artifact.

### Open

1. 🟡 **`PKG_UPGRADE` is untested and unprovenanced** — `Makefile:34` (`[ "${PKG_UPGRADE:-}" = 1 ] && exit 0`) is the *only* occurrence of `PKG_UPGRADE` in the repository: no test sets it, no script sets it, and nothing cites which package manager exports it. Either add a matrix case that runs the hook with `PKG_UPGRADE=1` (expect: skip) or cite the upstream convention that sets it. As written, the ledger line "host coverage models … upgrade/empty/unknown fail-closed paths" reads as if this path were pinned when it is not.

2. 🟡 **The lifecycle test never runs in CI with an artifact requirement.** `.github/workflows/fwlive-test.yml:101-105` runs only `tests/fwlive-package-payload.test.sh` under `FWLIVE_REQUIRE_PACKAGE=1`; `tests/fwlive-package-lifecycle.test.sh` is wired in at `scripts/fwlive-test.sh:195` *without* that env. Its artifact-inspection branch (`tests/fwlive-package-lifecycle.test.sh:35-65`) and its "no current package artifact found" failure path are therefore unreachable in CI — the CI-visible half of the new test is the action matrix only.

3. 🟢 **The artifact half is still a grep, not an execution.** `tests/fwlive-package-lifecycle.test.sh:52-53` asserts `default_prerm $0 $@` and `restore_wan_log_baseline` exist in the packaged files; a packaged body reverted to always-restore would still satisfy both. The action matrix does discriminate now, but only against the Makefile-extracted text (`:20-28`). Executing the IPK's `prerm-pkg` directly (or diffing it against the extracted body) would close this properly.

4. 🟢 **The version arm is a glob, not a version test.** `Makefile:42` `[0-9]*.[0-9]*` matches any digit-anything-anything-digit string (`1a2`, `12-`), so those also select the restore path. Real version strings are unaffected; a dotted-version pattern removes the ambiguity. (Unchanged from the pre-fix low.)

5. ⚠️ **Confirm build provenance of the recorded packages.** The pre-fix round examined the on-disk artifacts and reported the hashed x86 IPK's `prerm-pkg` used `[0-9]*)` with no inner fail-closed `*)` — i.e. built before this fix. There is no `out/` in the review checkout, so I could not re-verify. If those packages predate `78b984c5b4`, the x86 and APK lab cells exercised the **old** hook, and the hashes should be re-pinned from a rebuild.

Nothing else in this diff touches an ACL, DOM sink, or read/write scope, and the control row was honestly downgraded from `Host + lab` to `Host` here (re-promoted in #406 only once cells exist) — good bookkeeping.

lucas-albers-lz4:
Follow-up on the Composer/Grok/Bugbot + later verified findings (fixes in worktrees; main tree left for the security audit).

**Fixed**
- Lifecycle oracle now **executes** extracted `prerm-pkg`. Synthetic always-restore and old `$1=remove` IPKs fail; current body passes.
- `PKG_UPGRADE=1` matrix case + Makefile citation of OpenWrt `package-pack.mk`.
- Version arm requires a literal dot (`[0-9]*[.][0-9]*`); `1a2` must not restore.
- `default_prerm` `$1-pkg` hop for remove + upgrade.
- CI payload job runs `tests/fwlive-package-lifecycle.test.sh` with the same `FWLIVE_REQUIRE_PACKAGE` / `FWLIVE_PACKAGE` env.

**Skipped (intentional)**
- Rebuild/re-pin of hashed IPK/APK artifacts (no SDK run this pass).
- Live 0.1.44 → bumped-version QEMU upgrade cell. Makefile now states a version-changing APK upgrade is indistinguishable from uninstall at `pre-deinstall`.
- APK control scripts still skipped when `apk extract` has no `pre-deinstall` (scripts live in ADB metadata). IPK execution is required.

Commits: `7bf7c91996`, `8e18956c29`.

lucas-albers-lz4 (commented):
## Follow-up review — #405 @ `b8626abf6d`

Three independent reviewers (qwen3.7-max, minimax-m3, kimi-k2.7-code) plus a synthesizer, run against the refreshed heads because you pushed mid-wait. **Verdict: APPROVE** (3/3 agree the packaged hook discriminates).

I re-verified each of my earlier items against this head before writing:

| Earlier finding | Status now | Evidence |
|---|---|---|
| `PKG_UPGRADE` guard untested | **Fixed** | `tests/fwlive-package-lifecycle.test.sh:198-208` runs the hook with `PKG_UPGRADE=1` and asserts no restore |
| Lifecycle test never ran in CI with an artifact requirement | **Fixed** | `.github/workflows/fwlive-test.yml:106-110` now runs it under `FWLIVE_REQUIRE_PACKAGE=1` |
| Artifact half was a grep, not an execution | **Fixed (IPK)** | `oracle_ipk()` (`:260-303`) executes the extracted body and the real IPK's `prerm-pkg` (`:402`), with negative oracles that make an always-restore or old `$1=remove` body **fail** |
| `[0-9]*.[0-9]*` matched non-versions | **Fixed** | now `[0-9]*[.][0-9]*` with the `1a2` negative case (`:178-180`) |

The new `## Released v0.1.44 hook mismatch` section in the evidence record is the strongest part of the chain — it names the actual shipped defect (`[ "$1" = "remove" ]` never firing through the opkg wrapper, so restoration silently never happened) instead of leaving it implicit.

### Remaining (all Minor, non-blocking)

1. **The `Makefile:32-40` comment is now the only inaccurate statement left.** It says "A version-changing APK upgrade is indistinguishable from uninstall at this hook (both pass the old version as `$1`)". That is not how apk works: on upgrade apk runs `pre-upgrade`/`post-upgrade` and never `pre-deinstall` (Alpine's APKBUILD script set is pre/post-install, pre/post-upgrade, pre/post-deinstall; OpenWrt's `package-pack.mk` inlines the prerm body into `pre-deinstall` only). So the body is simply not invoked on upgrade, and the `PKG_UPGRADE=1` guard is future-proofing rather than a live branch. Suggest rewording to that effect — as written it implies a heuristic is all that stands between an upgrade and a silent WAN-logging reset, which is not the case.
2. **APK dispatch is instruction-checked, not executed.** `inspect_apk_artifact` (`:419-442`) greps the payload for `PKG_UPGRADE` / `case "${2-}"` / `remove`; a discovered APK hook whose `upgrade` arm fell through to `restore_wan_log_baseline` would still pass. Note also that an `apk extract` payload cannot carry control scripts, so that check will typically fall through to its note in the payload job — the installed `apk del` cell is what actually proves APK behaviour. A synthesized APK-shaped shim (`default_prerm` + inlined body, `$1=version`, `$2=""`) through the same matrix would close the asymmetry.
3. **Nits:** `run_opkg_hook` passes a `$WORK/usr/lib/opkg/info/` path before that directory is created (harmless — the path is pattern-matched, never opened, but ordering-fragile); `pack_synthetic_ipk` silently falls back to `tar -cf` when `ar` is absent.

### For the record — one reviewer finding refuted

Reviewer A raised a Major that a version-changing APK upgrade could restore the baseline (reasoning that `PKG_UPGRADE` is only exported by `post-upgrade`). **Refuted**: apk's purging branch is the only path that runs `pre-deinstall`, so the hook body is not invoked on upgrade at all and the baseline is preserved structurally. The valid kernel of that finding is item 1 above — the comment, not the code.

